PR #271, opened 2015-04-02
Feat bugfix 26.02.15

fkleedorfer:
Automatic merging isn't possible.
@sbyim are these fixes still valid?


sbyim:
these are still valid commits, i will try to merge with devel.

fkleedorfer:
dropped

PR #272, opened 2015-04-02
Topnav as cstm directive

fkleedorfer:
@m0ru The branch is not merged with devel, automatic merge is impossible
Are these fixes still valid? Should we merge?


fkleedorfer:
@m0ru shall we drop this?

fkleedorfer:
dropped

PR #273, opened 2015-04-02
Merged branch from gui-devel and feat-security-msg extended by some feat...

fkleedorfer:
Not merged with devel and conflicting. Do we need these changes?


fkleedorfer:
I think we don't need these changes anymore

PR #274, opened 2015-04-02
'New messages' as directive. Requires testing. Also fixed z-index for bo...

fkleedorfer:
@m0ru not merged with devel and conflicting. Is this still valid?


fkleedorfer:
@m0ru shall we drop this?

fkleedorfer:
dropped

PR #276, opened 2015-04-02
Fixing login

fkleedorfer:
@m0ru conflicting and not merged with devel. Do we still need this?


fkleedorfer:
@m0ru shall we drop this?

fkleedorfer:
dropped

PR #277, opened 2015-04-02
added verifier classes and basic implementations

fkleedorfer:
@yanapanchenko Not sure if we should merge this, please check


yanapanchenko:
It should not be merged at this state into devel, but it could be a continued for implementing the verifier used by WellformednessCheckingWonMessageProcessor

fkleedorfer:
@yanapanchenko now that we're working on the wellformedness check - can we re-use or close this?

PR #282, opened 2015-04-09
Feat fix and update for android app

quasarchimaere:


quasarchimaere:
yes, this will be replaced by parameterize-sparql-strings in the future
see: https://jena.apache.org/documentation/query/parameterized-sparql-strings.html

PR #289, opened 2015-04-27
Feat uri check processors

yanapanchenko:


yanapanchenko:
mistake - not for master

PR #291, opened 2015-04-28
Feat crawl for query

fkleedorfer:
Various changes in reaction to the changes that were made to the message format (now, whole graphs are copied, instead of the data, which results in a message with 2 URIs - a remote and a local one).

In general the changes are:
* the message timestamps are now called sentTimestamp and receivedTimestamp
* the owner-webapp's linked-data-service.js was rewritten to use sparql queries that have property paths attached, which allows crawling of the data before executing the query. Also, some of the service methods were rewritten to use this new approach.
* there is an EchoBot now, which makes testing easier: instead of running the matcher webapp and the MatcherBot, just run the EchoBot. It will connect to every new need and echo all text messages.

Sorry: accidentally committed most of the work in one commit with misleading message.


yanapanchenko:
I had the following problems with GUI so far:
- the need-x connection page with a match to need-y does not get refreshed when a connect request is received by need-x (sent from need-y)
- sometimes, after sending the text message, the last and all previous text messages are doubled, after receiving a text message it gets back to normal
- error when deactivating (closing) a post

PR #295, opened 2015-05-12
Matching service node controller setup

heikofriedrich:
setup node controller and do a first integration with crawler


heikofriedrich:
created new pull request #303

PR #336, opened 2015-08-25
Improve logger config

heikofriedrich:
kept the logback.xml now for local custom logging config


heikofriedrich:
@yanapanchenko please merge with devel if ok

PR #341, opened 2015-09-21
Minor bugfix

yanapanchenko:
Hints not displayed bugfix.

Note: another reason for hints not displayed is when the hint has matched need that is under the same user account. This is intended. Therefore, when testing current matcher, use different account (or private user) after each published need.


fkleedorfer:
@yanapanchenko do we still need this?

yanapanchenko:
no, it's already there

PR #345, opened 2015-10-21
Siren matcher integration

heikofriedrich:
Matching Service integration with siren matcher including bot load testing code and performance improvements for load test


fkleedorfer:
Change request: in module won-matcher-service 
* introduce base package `won.matcher.service` and move packages `common, node and crawler` there. 
* rename package `won.matcher.service.node` to something clearer, like `nodemanager`, `nodetracking`

PR #353, opened 2015-11-26
Architecture sketches

m0ru:
We started implementing the login and setting up the websocket-agent (currently it uses a mocked connection to echo messages). Also, I've split up actions.js and reducers.js for better readability (which is also why I'm making this pull-request before the agent is finished -> just leave the branch open after merging into devel.)


m0ru:
still under development

PR #359, opened 2015-12-17
Implemented Frontend Login/Register Form Validation

quasarchimaere:
-added register validation (angular form validation)
-added login validation (angular form validation)
-added password compare directive (compareTo)
-only enable form buttons when form is valid
-added generic disabled button style (lightgray bg + white fontcolor)


quasarchimaere:
closed because of additional changes

PR #362, opened 2015-12-18
Feat register enhancement

quasarchimaere:
Added links to landingpage when clicking signup buttons
+autoscroll to signup element (with focus on the email adress element)
+overlay to make it easier to see that the focus is on the register elements

Refactoring:
ng-class shorthand instead of one lined ifs


quasarchimaere:
tested on integration_test docker instance, merging the pull request now

PR #373, opened 2016-01-11
Feat linkeddata 2

sbyim:


quasarchimaere:
currently still fails because of some undefined methods/vars within feed.js (see $ngRedux which is not injected in the controller, furthermore there is some ngRedux state.getIn with an empty array which i do not understand -> should this be implemented already, because the mocked data is still in the source of feed.js

PR #387, opened 2016-01-20
Feat incoming requests

sbyim:


quasarchimaere:
There seems to be a Problem with the showing of Posts, the Posts are not visible after creation anymore, it takes another pageload to show them. Also the Matches are not present for already generated Posts (after a PageReload and after another Login), Logout does not reset the data yet, after the logout posts and matches are still present in the corresponding views -> Please fix these issues before this pull request can be merged

sbyim:
Following issues are now resolved.
- Logout resets Posts & Matches
- Posts get displayed after creation

PR #388, opened 2016-01-21
Feat feedback gui

quasarchimaere:
Added The GUI Elements in Matches (Flow View, Grid View) that Open a Send Request Dialog -> to see the Feedback Buttons you need to click on your Post within a Match (Bottom of the element), this should open a Dialog to send a request. Be aware that there is no backend integration for this yet and these are only Interface Elements without any Functionality so far


quasarchimaere:
wrong branch

PR #389, opened 2016-01-21
Feat feedback gui

quasarchimaere:
Added The GUI Elements in Matches (Flow View, Grid View) that Open a Send Request Dialog -> to see the Feedback Buttons you need to click on your Post within a Match (Bottom of the element), this should open a Dialog to send a request. Be aware that there is no backend integration for this yet and these are only Interface Elements without any Functionality so far


m0ru:
@sbyim's update got pulled into devel so that matches & incoming requests now work. however automatic merging with this branch isn't possible thusly atm. Could you merge from devel once more? :)

quasarchimaere:
@m0ru @sbyim i will merge devel into this branch first thing tomorrow and resolve the conflicts, also i try to write some (more) comments just to understand what i was trying to do

PR #392, opened 2016-01-22
Feat incoming req 2

sbyim:


quasarchimaere:
Displays requests as it was supposed to, however we should think about how the counters should work, i think that some counters should not only include the unread items -> merging anyway for now

PR #396, opened 2016-01-25
Overview post links

m0ru:
Just some links. Also: found some views we haven't implemented yet.


quasarchimaere:
merged with devel (because of conflict) -> we needed to pick which site gets "opened" after the creation of a need, the devel branch opened the feed, and this branch opened the post-visitor view (the created post), so i chose the created post for now... my be subject to change in a later release depending on the usability

PR #397, opened 2016-01-26
Post-overview shorter, localized timestamps.

m0ru:
This pull request is mutually exclusive to #398. It's a design decision which of these we want.


m0ru:
Closed in favour of of #398

PR #428, opened 2016-02-23
Feat feed

m0ru:
old connections-object collection is renamed as Deprecated

i tried to keep the old state structure as much as possible

important lookups (you'll need these when writing selectors):
```
state.getIn(['needs', 'ownNeeds', needUri, 'hasConnections']) // -> connectionUris
state.getIn(['needs', 'theirNeeds', needUri, 'hasConnections']) // -> connectionUris
state.getIn(['events', 'events', eventUri, 'hasReceiver']) // -> connectionUri
state.getIn(['events', 'events', eventUri, 'hasReceiverNeed']) // -> needUri
state.getIn(['connections', 'connections', connectionUri, 'belongsToNeed']) // -> needUri
state.getIn(['connections', 'connections', connectionUri, 'hasEvents']) // -> eventUris
```
when using `unreadEvents` try to just use the `.uri`-field and lookup the actual event in `['events', 'events']`

I still need to remove anything using `connectionsDeprecated` and remove the superfluos load-actions. but the state is in the form it can remain in for now, imo.

As usual, don't delete the branch after merging, plz :)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/account-actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/account-actions.js
@@ -17,7 +17,8 @@ export function accountVerifyLogin() {
             /* handle data, dispatch actions */
             .then(data => {
                 dispatch(actionCreators.user__loggedIn({loggedIn: true, email: data.username}));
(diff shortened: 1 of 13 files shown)


fkleedorfer:
Note: latest devel wasn't merged to this prior to the PR

PR #430, opened 2016-02-25
Feat accept request

quasarchimaere:


m0ru:
i merged into feat-feed. we can pull into devel once either of those two is finished

PR #437, opened 2016-03-09
Fix inputfield in ff

### LINKED ISSUE #414: Posting doesn't work in FF
During the posting process, something fails in firefox. The need is created on the node, a success response is received on the client and the redirect works as well. however the need isn't put into the list of owned needs.

Processing dies somewhere between the log messages "got response for CREATE:..." and "Dispatching action..." in [actions.js:228 to :245](https://github.com/researchstudio-sat/webofneeds/blob/feat-ld-paging/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js#L228)

m0ru:
de facto fixes #414


yanapanchenko:
Deployed in integration servers and checked in firefox: cursor is not jumping but the published needs,a dn their messages, are not displayed. In crome those created both in firefox and in chrome are displayed.

PR #438, opened 2016-03-11
Feat msg monitored

yanapanchenko:
Conversation-bot-based possibility to monitor and record messaging lifecycle steps time: from being generated to being sent, or to being read at reciver need, or to success responses being read by the message originator.


m0ru:
don't delete branch afterwards

yanapanchenko:
update: after merging, branch can be deleted

PR #441, opened 2016-03-15
Fix consistent events - accepting incoming requests working

m0ru:
There was a bug that―for incoming connection requests―`unreadEvents[cnctUri].uri` didn’t match any uri in the list of own events. The message handling was just using the first successful framing result―essentially yielding the remote counterpart of the own event, which of course had a different uri. A selector function in the feed-view was crashing as a result, while trying to access properties of an undefined event.

The code is now returning the correct event. What’s more, all events now also contain have property with their counterpart attached, as you can see in the state. Especially for incoming requests, this counterpart holds valuable information.


yanapanchenko:
When additional text message is provided it is not included in the sent message. Otherwise Accept works - open message generated and sent and gui reacts.

Issue #444, opened 2016-03-18
RDF for need with related connections and events


yanapanchenko:
For connections of a need container, we want to have a parameter, e.g. 'deeper', that would trigger construction of RDF that contains connections RDF, local need RDF of those connections, events of those connections. This should also work with paging, e.g. for a need include only last N connection with each of connections' last N events.

PR #448, opened 2016-03-22
Feat deeper ld

yanapanchenko:
Implemented #444 in a following way:

extended with connections and events need data is accessed via separate URL, - need resource URI plus 'deep' path fragment. Optionally, so that limited number of events and connections are returned, one can specify the size of the connections/events layer.

Example URL:
	https://example.com:8443/won/resource/need/1693454629993447400/deep?layer-size=5
This returns need resource description, plus 5 (latest by their events) connections resource descriptions, plus for each connection 5 latest events resource descriptions.
	
From the Owner client-side the requests should be done via the owner server linked data bridge (because of the webid proving - the need deep data includes confidential events data), for example:

https://example.com:8443/owner/rest/linked-data/?uri=https://example.com:8443/won/resource/need/1693454629993447400/deep?layer-size=5&requester=https://example.com:8443/won/resource/need/1693454629993447400


m0ru:
Awesome :D

PR #451, opened 2016-03-23
Feat deepld clientside

m0ru:
With this all rdf-resources get fetched in one go at page startup. There's only one caveats though: As we don't have the uris of the connections and events from the outset, we can't set their cache-state to pending or lock requests to them. So there's the possibility, that up to 3 request-slots are blocked with unnecessary requests that are subsumed by the deep triple set. We'd need to either lock the whole store or make the locking mechanism aware of our data-structure (so it doesn't lock other needs), to even partially address this.


m0ru:
I don't notice a noticeable speedup though (while running everything locally). There's still a lot of events, needs and connections beingt requested, that would either be the counterparts―but i haven't made sure of that yet―or pointless requests due to the foregone locking.

m0ru:
Now that caching is actually working there's a noticeable speedup (at 10 posts with 1 request and match each, owner and node running locally): 7.8s vs 12.8s
However the jsonld we get from the node contains a small mistake that prevents even further caching―the event and connection container's URIs don't end with slashes when they’re used as predicate. I’ve the response-json here: <https://gist.github.com/m0ru/31513f9e8c3d731794e6ead11bc43f35>

PR #452, opened 2016-03-24
Master depolyment

heikofriedrich:
automatic proxy nginx proxy configuration and master deployment with certificates


heikofriedrich:
@yanapanchenko 
i checked why it was not working last time you tried, i restarted the jenkins with "CREATE_CERTIFICATES_NEW" flag set and then it empties the database and it was working, before there was again stale data about the old certificates in the db 

=> i merged this branch

PR #460, opened 2016-04-07
added light buttons to indicate the presence of matches/requests/conv…

quasarchimaere:
…ersations even though there are no new ones, this just adds the icons and does not display them, displaying them when needed will be pulled with another branch in the future


quasarchimaere:
WRONG BRANCH

PR #466, opened 2016-04-11
Refac matches popup

quasarchimaere:
used the method introduced by @m0ru to determine whether a popup of a match (to send a request) will be displayed or not


quasarchimaere:
Also changed the old two way binding method for sentRequests and incomingRequests

PR #473, opened 2016-04-14
Fix refactoring ownertitlebar and connectionselect

m0ru:
A bunch of refactorings to reduce the API-surface of these two components, thus making them more stable and maintainable. This has the nice side-benefit, that you can reload the page on /post/owner/conversations atm without.

Also some other stuff I refactored while debugging for  #466


m0ru:
Probably needs some merging, as some of the code is duplicate to what #470 does

PR #488, opened 2016-04-18
Instant matcher test

fkleedorfer:
@heikofriedrich do we still need this branch?


fkleedorfer:
oops - should have been against devel.

PR #489, opened 2016-04-18
Instant matcher test

fkleedorfer:
@heikofriedrich do we still need this?


heikofriedrich:
changes already merged into devel => closed

PR #529, opened 2016-05-03
Refac unified ownerview

m0ru:
All owner views are reachable under `/post/` with this patch. The matches view still needs to be abstracted into a directive so it can be used in the matches tab. I've opened the pull request already so the routing-changes can be used for other branches.

#531 also makes sure the matches-tab works.


quasarchimaere:
problems that are (still) present:

- Closing the view of an open request within the post view (via clicking X) links to the main request view 
- view of an open request is also experiencing a header padding that is too big  (class: post-messages padding-top 2rem)
- post info view is not centered with the other content anymore (width is full page width alignment is left)
- post view can only be accessed while the user is logged in (wasn't the reason why we unify this view to allow anonymous access to needs via an url?)

m0ru:
This refactoring is a prerequisite step for having the public links. I've opened the pull-request early, so merging in the rather heavy refactoring doesn't become a problem later on ^__^

m0ru:
The padding and alignment bugs should be fixed now.

PR #543, opened 2016-05-18
Feat simple access control

m0ru:


m0ru:
I added some changes later in `fix_visitorsviewloading` that would take more effort to untangle than it's worth. Just merge that branch. :)

PR #544, opened 2016-05-18
Fix visitorsviewloading + simple access managment

m0ru:
When you pasted a uri for the post-view while the page already was loaded (thus after the `initialPageLoad` action was created and dispatched) the need wouldn't get loaded. The assumption before was that pasting and entering always triggers a page-reload. However when the base-url matches, the browser only "jumps" to the new fragment-identifier, without reloading and thus without triggering `initialPageLoad`. 

This patch adds a listener to route-changes and causes the post to be loaded if the user would run into an unitialized post-view.

This branch also include simple access managment.

Check the following behaviours:

* once logged in and once logged out do:
  * reload on posts-overview
  * while on some other view (e.g. landing-page): paste link to posts-overview
  * reload on post-detail view
  * while on some other view (e.g. landing-page):paste link to post-detail view
* log-out on posts-overview
* log-out on posts-view (probably still redirects incorrectly)

you should always have data in those views and when you try accessing the overview while logged out you should land on the last page you were on or the landing page if this is the first in the app you accessed.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
@@ -138,7 +138,8 @@ const actionHierarchy = {
     router: {
         stateGo,
         stateReload,
-        stateTransitionTo
+        stateTransitionTo,
+        accessedNonLoadedPost: INJ_DEFAULT, //dispatched in configRouting.js
     },
     posts:{
         load:INJ_DEFAULT,
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
@@ -139,6 +139,7 @@ const actionHierarchy = {
         stateGo,
         stateReload,
         stateTransitionTo,
+        back: stateBack,
         accessedNonLoadedPost: INJ_DEFAULT, //dispatched in configRouting.js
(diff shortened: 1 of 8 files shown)


quasarchimaere:
if you sign up from the landingpage without ever been signed in (with any other user) the routing will not be changed (there is a reroute loop) 

See follow Log-Snippet:

Not sure about login-status -- will reinitialize route-change once it's been clarified.
configRouting.js:193 Tried to access view that won't workwithout logging in. Blocking route-change. Error: transition prevented(…)(anonymous function) @ configRouting.js:193
configRedux.js:10 action:   Object {type: "@@reduxUiRouter/$stateChangeStart", payload: Object}
configRedux.js:11 changing state from  Object {config: Object, router: Object, user: Object, messages: Object, lastUpdateTime: 1463985830774…}
configRedux.js:15 changed state to  Object {config: Object, router: Object, user: Object, messages: Object, lastUpdateTime: 1463985830780…}
configRouting.js:180 Not sure about login-status -- will reinitialize route-change once it's been clarified.
configRouting.js:193 Tried to access view that won't workwithout logging in. Blocking route-change. Error: transition prevented(…)(anonymous function) @ configRouting.js:193
configRedux.js:10 action:   Object {type: "@@reduxUiRouter/$stateChangeStart", payload: Object}

m0ru:
\*Fixed\* :)
(see detailed commit-msg)

Issue #566, opened 2016-05-24
Received requests and matches don't show up


m0ru:
The receiving app displays a lot of console output that suggests, that the incoming request is received and the counterpart-need's data downloaded. However, the displayed gui doesn't reflect this. Either the reducer doesn't handle the action or none is dispatched.

PR #578, opened 2016-05-25
Fix receiverequests #566

m0ru:


quasarchimaere:
an old bug is present again:

if you sign up from the landingpage without ever been signed in (with any other user) the routing will not be changed (there is a reroute loop)

PR #579, opened 2016-05-25
Feat send close conn to owner on deactivate need

fkleedorfer:
deactivate ('close need' in the owner webapp) causes a close message for each connection of the need to be sent. 

When closing a need, its connections should be gone upon re-opening. 
However, the GUI doesn't handle the closing correctly, yet (it seems), we may continue development of this feature in this branch until the GUI reacts properly


fkleedorfer:
Note: because of a change to WellformednessCheckingWonMessageProcessor, message processing should now be considerably faster.

m0ru:
Messages don't arrive at the browser yet. @fkleedorfer is looking into it.

fkleedorfer:
Should work now. Running on integration-test

PR #608, opened 2016-05-31
Fix lostrequests 598

m0ru:


fkleedorfer:
The duplicate posts issue still not completely resolved, the submit button sometimes stays deactivated now. However, it's not possible to double-post something.

PR #616, opened 2016-06-03
Incrementally dispatching during page-load, when signed-in.

### LINKED ISSUE #575: Incremental dispatch at initialPageLoad
As this action-creator takes quite a while, we could dispatch the action every time any part of the data comes in, causing the gui to incrementally update, projecting the appearance of increased performance.

m0ru:
Resolves #575


fkleedorfer:
I don't see any difference in behaviour.. page load is still monolithic both in FF and Chrome

fkleedorfer:
I would like to merge this if someone can convince me it's better than master.

m0ru:
Basically you should see a fully rendered posts-overview before all the events have loaded. However I tried it a few times now and the speed-up is neglible. We should probably just close this pull-request and work on loading the data jit on route-changes (whatever's needed for the current view). The feed will probably be rather tricky for that one, though.

PR #619, opened 2016-06-03
Feat removedummies #587

### LINKED ISSUE #587: Remove Dummy Images & Dummy Texts
Remove all the dummy and fake images for posts and dummy texts (locations timespans etc.)

m0ru:
Check:

* post::info
* matches::gridview
* matches::*::sent-request

Fixes #587


fkleedorfer:
Regression bug: 
When posting, the behaviour described in #414 is back again: in FF, the cursor jumps to the beginning of the text field repeatedly while typing, in Chrome all is fine.

m0ru:
Yeah, contenteditable seems to be not very standardized. Every browser generates different html (e.g. FF generates a trailing `<br>`)
I'll change the textfield to use a micro-version of our architecture within itself:

1. Keypress events will be directly registered and swallowed before they reach the contenteditable
2. these update the internal model: an array of strings (=lines of text) plus the caret-position
3. the resulting model gets rendered as html and the caret position set 

This should avoid any more regression bugs for the textfield. As an added bonus, should we ever allow richer formatting, this should be a lot simpler to maintain.

Also see a blog-post by a developer for medium.com on the topic: [Why ContentEditable is Terrible,
Or: How the Medium Editor Works](https://medium.engineering/why-contenteditable-is-terrible-122d8a40e480#.2ut5hxm1w]

fkleedorfer:
Can we merge this? If so, does this version of dynamic-textfield.js supersede the one in master?

m0ru:
Now we can :)

PR #660, opened 2016-07-04
cleanup remove old matcher code and update doku

heikofriedrich:
- remove old matcher node and doku about matching
- fixes #660


fkleedorfer:
@heikofriedrich please leave a message here when the PR is ready for review

heikofriedrich:
@fkleedorfer can be merged now

webofneeds/won-matcher/src/main/java/won/matcher/component/MatcherNodeURISource.java
```diff
@@ -1,31 +0,0 @@
...
- */
-
-package won.matcher.component;
-
-import java.util.Iterator;
-
-/**
- * Interface that is used to obtain a WON_BA node URI.
- */
-public interface MatcherNodeURISource
```
fkleedorfer: We cannot delete this interface and its implementations: they are still used. I am sceptical if we should really delete won-matcher at all.. The bots will not work without them.

heikofriedrich:
ok, i didnt see this when i did the build, shall we keep the won-matcher project and anyway delete won-matcher-solr and won-matcher-webapp?

heikofriedrich:
@fkleedorfer didnt delte the won-matcher project => can be treviewed

fkleedorfer:
I see - but now we deleted won-matcher and then added the code again as new, which means that we lose the whole edit history of that code. Not ideal.

heikofriedrich:
ah ok, i will fix that

heikofriedrich:
@fkleedorfer fixed

Issue #631, opened 2016-06-14
unread counters do not count down


fkleedorfer:
Create a need and wait for the debug bot to send hint and connect. The unread counters in the overview will count up but when view the new connections (or even accept or decline), the counters do not count down.

m0ru:
Currently the events aren't marked as read at all, so they just stay in the list of unreadEvents till the next page-reload.

I think it should be possible to make the event-reducer update the state during the state-go actions, in particular for the following views:

* incoming request: entirely folded out with conversation on the right side
* messages: entirely folded out with conversation on the right side
* in matches-list: an entry entirely opened (the popup is visible)
* matches grid/tiles-view?

The process should be the same for the overview and non-overview variants. Also, all unread events for the connection should be marked as read (unless we want to factor in how far down people have scrolled in a conversation).

fkleedorfer:
partly done in #744
however: these are still TBD
* does not count down matches when I click the matches tab. (may be ok)
* does not count down matches when I decline a match (but I feel it should)
* does not count down messages for a need when I open the conversation view
* counts messages up while I receive them in the conversation view
* does not count down requests when I click 'requests' and accept a request.

quasarchimaere:
implemented -> no longer relevant

PR #672, opened 2016-07-06
Fixed the counters @ overview-titlebar and post-overview

m0ru:
(doesn't fix the more work-intensive #631 though, as there's higher-priority issues that we should focus on first)


fkleedorfer:
ok so what is the PR supposed to fix exactly?

fkleedorfer:
Ok, got it: 
* counters count up correctly now in 'Posts' view and in the overview titlebar
* counters do not count down yet

fkleedorfer:
oops - didn't want to close!

PR #673, opened 2016-07-06
Fix layout issues with long titles

### LINKED ISSUE #570: Layout problem for long titles
This title isn't even that long—we allow got up to 140 characters(!)—however the layout already breaks horribly:

not enough space for title

Either we need a way that it can continue over the tab-titles or start with the responsive layout at relatively wide viewport widths. This is a good example for why mobile-first is a better approach when designing. It's always easy a low-effort option to do something with the extra-space when you go from small viewport to large viewport, but things quickly break design-wise and cause a lot of unavoidable work the other way around.

### LINKED ISSUE #674: Layout problem with long titles in post-overview
screenshot of broken layout

solvable by setting the row to `flex-wrap: nowrap`

related to #673

m0ru:
Resolves #570 
Resolves #674


m0ru:
You can delete the other `fix_titlelayout_*`-branches as well after merging this.

PR #675, opened 2016-07-07
add in memory hint duplicate filter

### LINKED ISSUE #663: Implement Hint database for matcher
* currently only implementation that just filters duplicates 
* simple implementation as Bloomfilter (best guess duplicate detection) 
* that gets saved and load from file for persistence
* checked when hint is going to be sent
* updated if need hasnt been sent yet + send hint
* warning/info if hint is not sent (but could be a hint that hasnt been sent yet = false positive) => no further lookup in that situation for now

heikofriedrich:
fixes #663


fkleedorfer:
This will not do anything until we add more matchers that produce overlapping results. Only if we persist the bloom filter, we could test what happens if we restart the matching service. After a re-crawl, duplicate hints might be sent, with the filter, this would be prevented. 
So.. we cannot really do integration tests here

PR #678, opened 2016-07-08
Fix openmsg #651

### LINKED ISSUE #614: Sent request not moved to messages if accepted
When the other party accepts a sent request, it's not moved to messages yet (a page-reload is necessary to achieve that)

### LINKED ISSUE #651: conversations tend to hang at some point
I feel this is reproducible every time although it is not every time exactly the same:

Work with 2 browsers and 2 accounts.
1. create a need in one account and one in the other
2. wait for the match, then connect
3. exchange messages

During this process, at least once, each webapp will stop receiving messages and you'll have to reload

m0ru:
The openMessages you get after a request of your's is accepted are now handled. To test, open needs in two accounts, make a request from the one, accept with the other. In the first account the connection should now move from the "Sent Requests"- to the "Messages"-tab.

Also, I refactored the action-creators for message handling to avoid some of the code-duplication and made the batch-rdf-loading via `urisToLookupMap` a bit more resilient against network-errors. The ressources that failed to won't show up in the result map, without breaking everything else.

Resolves #651 
Resolves #614


fkleedorfer:
Doesn't fix #651 - renamed that issue to point out that the app doesn't necessarily hang after connecting but also while receiving hints

PR #686, opened 2016-07-14
Feat toast messages

quasarchimaere:


fkleedorfer:
Cannot really test.
Tried this in the console
```
otb.$ngRedux.dispatch(actionCreators4Dbg.needs__failed({error:new Error("bla")}))
```
and the result isn't what I expected - no icon is shown

unbenannt

also, the toast is transparent, which doesn't look so good when it overlays something else:

unbenannt

quasarchimaere:
use: otb.$ngRedux.dispatch(actionCreators4Dbg.toasts__test({error:new Error("bla")})) to test all 3 different  toast types

PR #696, opened 2016-07-15
add synonym filtering to siren queries

### LINKED ISSUE #695: add synonym database to solr

heikofriedrich:
fixes #695


heikofriedrich:
i updated the solr docker image to use the synonyms now and pushed it, every new deployment of this solr server will have the synoyms included

PR #701, opened 2016-07-21
add Kneedle knee detection algorithm to matcher

heikofriedrich:
- just added class, usage will follow

### DIFF
--- a/webofneeds/won-matcher-siren/pom.xml
+++ b/webofneeds/won-matcher-siren/pom.xml
@@ -52,6 +52,16 @@
             <artifactId>jena-arq</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
+            <version>3.6.1</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


webofneeds/won-matcher-siren/pom.xml
```diff
@@ -52,6 +52,16 @@
             <artifactId>jena-arq</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
+            <version>3.6.1</version>
```
fkleedorfer: please add the dependency to the `<dependencyManagement>` element of the parent pom and remove the version info here

webofneeds/won-matcher-siren/src/test/java/won.matcher.siren.utils/KneedleTest.java
```diff
@@ -0,0 +1,44 @@
+package won.matcher.siren.utils;
```
fkleedorfer: something is wrong with the folder structure

PR #703, opened 2016-07-22
Started working on conditionally allowing line-breaks.

fkleedorfer:
@m0ru @quasarchimaere please check if we still need this one commit, it's from quite some time ago when @m0ru started work on descriptions


m0ru:
With descrtiptions they're allowed now and we don't need the component in any other context, so: nope, we don't.

PR #713, opened 2016-08-02
replaced Siren matcher with Solr matcher

### LINKED ISSUE #708: Get rid of siren
* Then for instance upgrade to Solr 6.1 => use spatial queries, and hopefully (?) working date fields
* currently i see no benefit in using siren anymore 
* we can also have a dynamic schema in pure Solr + flattening our JSON and search the fields individually
* it was not to be updated anymore since more than a year 
* a bit reading and testing should be done to decide if Solr is really applicable for our purpose
* http://lucene.apache.org/solr/quickstart.html

heikofriedrich:
fixes #708 
* matching should work basically the same as before, just without siren


heikofriedrich:
@fkleedorfer fixed the pom problems and pushed on integration test for testing, everything should be running now

fkleedorfer:
merging - please change the threshold to be applied to the solr score **before** normalization as discussed

Issue #18, opened 2014-07-07
owner and owner webapp: hint does not contain facets


fkleedorfer:
hint object only contains needURIs, not facet URIs, so a default facet has to be used. Should be changed such that hints are no longer used, but a hint method (or object) is used that refers to the connection object that was created on the node in reaction to the hint.

Issue #14, opened 2014-06-10
some restrictions for connection between different facet types might be needed


sbyim:
eg) CommentFacet -> connect -> OwnerFacet

PR #717, opened 2016-08-04
Feat fewer envelopes

fkleedorfer:
How to test: deploy and see if everything (including closing connections from the remote side) still works

### DIFF
--- a/webofneeds/won-cryptography/src/main/java/won/cryptography/rdfsign/WonVerifier.java
+++ b/webofneeds/won-cryptography/src/main/java/won/cryptography/rdfsign/WonVerifier.java
@@ -7,6 +7,8 @@ import com.hp.hpl.jena.rdf.model.StmtIterator;
 import de.uni_koblenz.aggrimm.icp.crypto.sign.algorithm.SignatureAlgorithmInterface;
 import de.uni_koblenz.aggrimm.icp.crypto.sign.algorithm.algorithm.SignatureAlgorithmFisteus2010;
 import de.uni_koblenz.aggrimm.icp.crypto.sign.graph.GraphCollection;
+import org.apache.jena.riot.Lang;
+import org.apache.jena.riot.RDFDataMgr;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -15,6 +17,7 @@ import won.protocol.util.RdfUtils;
 import won.protocol.util.WonRdfUtils;
 import won.protocol.vocabulary.WONMSG;
 
+import java.io.StringWriter;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.Provider;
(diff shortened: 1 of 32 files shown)


fkleedorfer:
Open Todos:
* update message-signing.md to reflect changes
* put public key in connect/open message so that we have a signed message containing that key on our side (in the owner application). 
* check for the protocol version identifier, only allow 1.0

webofneeds/won-node/src/main/java/won/node/camel/processor/general/ReferenceToUnreferencedMessageAddingProcessor.java
```diff
@@ -95,7 +97,7 @@ public WonMessage process(final WonMessage message) throws WonMessageProcessingE
     }
     //persist the message
     messageEventRepository.save(messageEventPlaceholders);
-    return new WonMessage(messageDataset);
+    return message;
```
heikofriedrich: is it save to change the message object and return it or should the changes be made on a that is returned copy?

webofneeds/won-cryptography/src/main/java/won/cryptography/rdfsign/WonVerifier.java
```diff
@@ -15,6 +17,7 @@
 import won.protocol.util.WonRdfUtils;
 import won.protocol.vocabulary.WONMSG;
 
+import java.io.StringWriter;
```
heikofriedrich: this message seems to print a warning now in the owner logs:

9.8.16 12:57:59.350 [Camel (wonOwnerCamel) thread #18 - JmsConsumer[OwnerProtocol.Out.257ec8ad1128a9f6d4f4a5ca77356fffa04d457b180f2762a7869d8f]] WARN  won.cryptography.rdfsign.WonVerifier - cannot verify signature https://satsrv06.researchstudio.at/won/resource/event/6631698687093502000#envelope-l6rb-sig as it is not part of this message
9.8.16 12:58:00.041 [Camel (wonOwnerCamel) thread #14 - JmsConsumer[OwnerProtocol.Out.257ec8ad1128a9f6d4f4a5ca77356fffa04d457b180f2762a7869d8f]] WARN  won.cryptography.rdfsign.WonVerifier - cannot verify signature https://satsrv06.researchstudio.at/won/resource/event/rr8eomkdztufirwvuhqm#envelope-xz3q-sig as it is not part of this message
...

i was creating two needs, got some matches and created a connection and exchaned a few messages. => Fix this warning messages

heikofriedrich:
@fkleedorfer reviewed the code, made some comments, tell me in a bit more detail how i can test the remote connection closing? can i use the gui? and how to verify then that the message was closed

heikofriedrich:
ok could test this a bit now and "close" seemed to close the connection, however after "deactivate" i could still write messages even though the bot was not responding anymore. is this right?

webofneeds/won-node/src/main/java/won/node/camel/processor/general/ReferenceToUnreferencedMessageAddingProcessor.java
fkleedorfer: This is intended, and it's not really safe. However, I gave up the concept of immutable WonMessage objects for the current implementation as it creates a big management overhead. This way, it became easy to add properties to a newly added envelope (see also changes to WonMessageRoutes) one by one, and be sure that only one envelope is added as the message passes through the system.

webofneeds/won-cryptography/src/main/java/won/cryptography/rdfsign/WonVerifier.java
fkleedorfer: The warning is there as a reminder. When a message references the signature of another message (which is not copied into the dataset), we would have to load that message's data in order to be able to verify the signature. This step is currently skipped.

fkleedorfer:
Please file an issue for that. I am pretty sure this bug is not new. I think that the messages are not accepted by the server, but in the GUI, we are processing them optimistically and probably fail to handle the FailureResponse.

heikofriedrich:
isues created #725

PR #729, opened 2016-08-10
Fix too many emails

### LINKED ISSUE #646: All 3 Url links in email do not work
when clicking on either of these three links in the email that is received upon coversation request, you are forwarded to the start page and do not see the specified need or conversation

### LINKED ISSUE #661: Too many conversation message emails
I created an account on tintegration test, created two need and chatted like 5 sentences with the bot => got 14 emails => 2 conversation requests + 12 coversation message emails, even though i already saw all the messages from the other node while chatting

fkleedorfer:
A few changes to the way the owner app sends emails:
* emails are only sent when no websocket session is found
* text was improved 
* links were improved

Not all the links work, though, as it is currently unclear how to link directly into a conversation - so this link doesn't work completely, but it gets you to the post view at least.

How to test:
* deploy to integration-test
* use two user accounts with real email addresses
* create two needs that will be matched, one in each account
* log out quickly with account A after creating the need
* check that you got the hint email of account A
* go to account B, request contact on the hint
* check that you got the open email in account A
* log in with A, accept the conversation, send a message
* reply with B then log out
* reply with A
* check that you got the message email for B
* close the connection with A
* check that you got the close email for B

fixes #661 
fixes #646

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/java/won/owner/web/WonOwnerMailSender.java
+++ b/webofneeds/won-owner-webapp/src/main/java/won/owner/web/WonOwnerMailSender.java
@@ -5,6 +5,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import won.owner.service.impl.URIService;
+import won.protocol.model.ConnectionState;
 import won.protocol.util.WonRdfUtils;
 import won.protocol.util.linkeddata.LinkedDataSource;
 import won.utils.mail.WonMailSender;
@@ -24,15 +25,16 @@ public class WonOwnerMailSender {
   @Autowired
   LinkedDataSource linkedDataSource;
 
-  private static final String OWNER_REMOTE_NEED_LINK = "/#/post/visitor/info/?theirUri=";
-  private static final String OWNER_CONNECTION_LINK = "/#/post/visitor/messages/?myUri=<>&theirUri=";
-  private static final String OWNER_LOCAL_NEED_LINK = "/#/post/owner/info?myUri=";
+  private static final String OWNER_REMOTE_NEED_LINK = "/#post/?postUri=";
+  private static final String OWNER_CONNECTION_LINK = "&connectionUri=";
+  private static final String OWNER_LOCAL_NEED_LINK = "/#/post/?postUri=%s&connectionType=%s";
 
   private static final String NOTIFICATION_END = "\n\n\n" +
     "Sincerely yours,\nOwner application team" +
     "\n\n\n\n" +
     "You have received this email because you are subscribed to getting " +
-    "notifications for your posting. If you received this email in error please click on the link below." +
+    "notifications for your posting. If you think we should not have sent you this email, please click on the link " +
+    "below." +
     "\n" +
     "\n\n" +
     "This is an automatic email, please do not reply." ;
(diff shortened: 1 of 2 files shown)


heikofriedrich:
* It is better now to receive less mails but i think there are still too much. I should not receive a email for every single text message someone sent me while im not logged in. 
* I didnt receive any hint message, maybe im not logging out fast enough but i think this way inteded not to receive emails for hint messages cause this would be even more spamming
=> anyhow i would merge this cause its an improvement of the situation

webofneeds/won-owner-webapp/src/main/java/won/owner/web/WonOwnerMailSender.java
```diff
@@ -24,15 +25,16 @@
...
   LinkedDataSource linkedDataSource;
 
-  private static final String OWNER_REMOTE_NEED_LINK = "/#/post/visitor/info/?theirUri=";
-  private static final String OWNER_CONNECTION_LINK = "/#/post/visitor/messages/?myUri=<>&theirUri=";
-  private static final String OWNER_LOCAL_NEED_LINK = "/#/post/owner/info?myUri=";
+  private static final String OWNER_REMOTE_NEED_LINK = "/#post/?postUri=";
+  private static final String OWNER_CONNECTION_LINK = "&connectionUri=";
+  private static final String OWNER_LOCAL_NEED_LINK = "/#/post/?postUri=%s&connectionType=%s";
 
   private static final String NOTIFICATION_END = "\n\n\n" +
```
heikofriedrich: This constant String is never used, is this intended? Shall we delete it?

webofneeds/won-owner-webapp/src/main/java/won/owner/web/WonOwnerMailSender.java
```diff
@@ -259,21 +168,61 @@ public void sendConnectNotificationHtmlMessage(String toEmail, String localNeed,
...
+        "   closed the connection to <a href=\"" + linkLocalNeed + "\">you</a>. " +
+        theyWrote +
+        "<p><a href=\"" + linkConnection + "\">[Click here to view the conversation]</a></p>" +
+        NOTIFICATION_END_HTML +
+        "</div>";
+
+
+    logger.debug("sending " + SUBJECT_CONNECT + " to " + toEmail);
+
+    this.wonMailSender.sendHtmlMessage(toEmail, SUBJECT_CONNECT, body);
```
heikofriedrich: this should be another subject for the closing of messages not "new conversation request" (SUBJECT_CONNECT)

webofneeds/won-owner-webapp/src/main/java/won/owner/web/websocket/WonWebSocketHandler.java
```diff
@@ -242,30 +245,37 @@ private void notifyPerEmail(final User user, final URI needUri, final WonMessage
...
           }
           return;
         case CONNECT:
           if (userNeed.isRequests()) {
             emailSender.sendConnectNotificationHtmlMessage(
-              user.getEmail(), needUri.toString(), wonMessage.getSenderNeedURI().toString(), textMsg);
+              user.getEmail(), needUri.toString(), wonMessage.getSenderNeedURI().toString(), wonMessage.getReceiverURI().toString(), textMsg);
           }
           return;
         case HINT_MESSAGE:
```
heikofriedrich: as i said in the comment earlier im not sure if we should keep this, however i think it does not work anyway.
I tested it again and i do not receive any hint messages when im logged out

heikofriedrich:
@fkleedorfer reviewed code and tested on int

webofneeds/won-owner-webapp/src/main/java/won/owner/web/WonOwnerMailSender.java
fkleedorfer: true, good catch

webofneeds/won-owner-webapp/src/main/java/won/owner/web/websocket/WonWebSocketHandler.java
fkleedorfer: yes, the default config is: no mail for hints

fkleedorfer:
@heikofriedrich your proposition to send even fewer messages (by using a daily digest, for example) is in line what was discussed with @yanapanchenko. It's another issue though.

fkleedorfer:
ok fixed the close-mail subject bug

fkleedorfer:
@heikofriedrich ready for testing again

Issue #688, opened 2016-07-14
Add/show Location


fkleedorfer:

m0ru:
comparing map tools:

* Google maps: this leads to people making requests to google's API on pageload, essentially undermining the privacy we're striving to offer.
* Bing maps: same issues as with Google maps.
* Open-layers: A very feature-rich / heavy-weight library that can render map material from different sources (e.g. openstreetmaps). We should be able to add map-data hosting to the owner server to remove the cross-origin queries for map material, entirely solving this privacy leak. For now, at least the data doesn' t go to Google.
* Leaflet: a slimmer alternative to open-layers
* There's a few non-free solutions as well that aren't an option for our project due to that reason.

m0ru:
As for the design: 
I think for the workflow it might be faster to prefill the "default address" and allow the user to change the location, e.g. by picking one of the previously entered addresses (extractable from `ownNeeds`) or searching for a new address.

m0ru:
![ui-design for location picker](https://cdn.rawgit.com/m0ru/c4bf85134156da705d35966b79d6c37c/raw/0d186a7a86b7d4ee9d06b1a369b372dcfe530453/location_picker.svg)

m0ru:
As of a64b50b the following TODOs are still open:
* styling
* current location
* previous locations
* fixing layer-select icon in leaflet
* placing markers for selection
* zooming to proper size around these
* displaying selection as text

m0ru:
As of f3a6914fdb06dd11a96516879af56286c13333ca current location works, markers get placed, the input veriefied and posted to the server. The following TODOs are open:

* styling, including but not limited to:
  * highlighting of the search string in results
  * won-themed map-markers
* previous locations
* fixing layer-select icon in leaflet
* deriving a zoom-level for the query from current-location's accuracy-measure
* toaster notification if getting the current location fails
* allowing people to enter and post addresses not recognised by nominatim

m0ru:
The location information of owned needs doesn't show up in the state, but is in the rdfstore. The culprit is the [sparql-query for needs](https://github.com/researchstudio-sat/webofneeds/blob/818fa31865417c4750d8082db54354ae4451cf3f/webofneeds/won-owner-webapp/src/main/webapp/app/service/linkeddata-service-won.js#L830) in the `linkeddata-service-won.js`. It assumes a flat-ish structure, where the location information is actually nested.

A need example:

```
 need:946717029499422700
            a                     won:Need ;
[...]

Issue #402, opened 2016-01-28
front-end: displayed order of text messages sometimes wrong


yanapanchenko:
Possible reason:
at least in the old GUI, when querying the text messages for displaying in conversation view, the messages are ordered by received timestamp of local message and of the corresponding remote message at the same time. If the messages are hosted at different nodes that have time differences, it might lead to the wrong ordering of messages.

yanapanchenko:
reletaed issues #304 and #299

fkleedorfer:
This is still an issue, reproducible in the integration-testing system. The ordering is consistent within each sender but inconsistent between senders.  As @yanapanchenko points out, the possible cause is that the two servers' clocks are not synchronized and ordering uses timestamps of both nodes (but they should only use the timestamps of the owner's node)

heikofriedrich:
![chat1](https://cloud.githubusercontent.com/assets/7766279/16228062/f8092884-37b4-11e6-80b3-8dff11ecfb0a.png)

Also experienced that, my message are sorted all to the top while the bots messages are added at the bottom while we talk ...

m0ru:
Might be caused by how `won.getEvent` is implemented. If it encounters multiple timestamps, it uses the latest (which is flawed if they're from different servers). It should always use the the timestamp from the own owner.

fkleedorfer:
A while ago, we wrote a SPARQL query that takes care of this for the mailbot, @quasarchimaere and I can help

fkleedorfer:
This is still an issue btw, however, it seems, only when the two needs are on different nodes.

m0ru:
We can use the `hasPreviousMessage` triples to get a rough order of messages without looking at the timestamps (and only use the latter as secondary sorting where messages where send at almost the same time and thus have the same previous and next message)

This will be easier to debug once #1092 is resolved. The daisy chaining is also needed for #1056.

m0ru:
https://github.com/researchstudio-sat/webofneeds/pull/1124/files#diff-56522a221b044535f52269b176b6d230 contains a sparql-query for retrieving the messages in a sorted fashion. This should give us an initial sorting; new messages received via the websocket would only need to be inserted at the right position. We'd have to rerun the query whenever "show more" is used though.

m0ru:
Ah, I forgot to run `mvn clean`. It's good now.

m0ru:
here we go again:

[...]

PR #732, opened 2016-08-10
Reworked need-structure in state

m0ru:
During the work on the location fields, it was necessary to rewrite the sparql-query that pulls the need-rdf. @fkleedorfer suggested to then use the json-ld structure directly in the state without transformation. This pull-request contains that structure update, the resulting refactorings and a minor bug-fixes that happened en passant. Please merge soon, so that the currently developped code uses the new structure. :)

The location picker is currently commented out. Please leave the branch open as I'll continue working on it. :)


m0ru:
Just found a few components that I had overlooked during refactoring.

m0ru:
Now, that should be it.

m0ru:
See #688 for documentation.

fkleedorfer:
#402 seems to be more of an issue now

fkleedorfer:
#402 seems to be more of an issue now

fkleedorfer:
On PostInfo, the map that is displayed is zoomed out maximally, to the whole world

fkleedorfer:
The feed doesn't show any items:
feed

m0ru:
Reopening doesn't seem to work either as the success-response isn't handled.

fkleedorfer:
During a conversation, when you click 'close connection', the conversation panel and the need (left side) disappear, then reappear after a few seconds. clicking 'close connection' again makes it diappear permanently

m0ru:
The [map bug](https://github.com/researchstudio-sat/webofneeds/pull/732#issuecomment-239412883) occurs when you switch to the post-info page from another page (instead of reloading there). Seems the watch-expression wasn't quite working as expected.

m0ru:
[Ad](https://github.com/researchstudio-sat/webofneeds/pull/732#issuecomment-239412634) #402: I wrote a new handler for selecting timestamps that looks at whether the own need is the receiver or sender and then goes on to pick the send- or received-time. I guess it doesn't work that way, does it? Probably we should change it to check for the node-uri. 

Or change the message-format so there's only one timestamp as we've talked about.

quasarchimaere:
last commit fixed the feed issue

PR #737, opened 2016-08-11
discard doNotMatch flagged needs in Solr matcher 

heikofriedrich:
* and use a test core for usedForTestingNeeds
* needs created will now only be matched with one debug need (cause the debug bot does this), but not with all other debug needs available anymore
* how to test: create a some needs and check that each one them is matched with only 1 debug need (the one created for it) and any number of normal needs

### DIFF
--- a/webofneeds/conf/matcher-solr/matcher-solr.properties
+++ b/webofneeds/conf/matcher-solr/matcher-solr.properties
@@ -15,7 +15,11 @@
 #
 
 # solr database uri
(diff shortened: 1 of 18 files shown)


fkleedorfer:
Does not seem to work - I don't get any hints at all (except the ones from the DebugBot)

heikofriedrich:
hmm ok ill have a look

heikofriedrich:
@fkleedorfer i fixed the problem, i was querying the wrong index (switching test and default index) i should work now, deployed on integration test

PR #744, opened 2016-08-12
Unread-events get marked as seen depending on routes visited.

m0ru:
While debugging/testing for #732 I noticed that the unread-counters weren't decreased. This patch should fix that. The event-reducer looks at the route currently displayed and marks all events belonging to displayed connections as seen.


fkleedorfer:
* does not count down matches when I click the matches tab. (may be ok)
* does not count down matches when I decline a match (but I feel it should)
* does not count down messages for a need when I open the conversation view
* counts messages up while I receive them in the conversation view
* does not count down requests when I click 'requests' and accept a request.

fkleedorfer:
needs merging now

Issue #623, opened 2016-06-10
Load data selectively


fkleedorfer:
Instead of loading everything  on startup, determine what needs to be loaded and load only that.

m0ru:
I think we should really break this down in seperate tasks and merge work on them into a `feat_load_selectively` branch.

 A breakdown of steps might be as follows:

1. Chat messages: only load when the connection is opened.
1. Chat messages: 
  * only load latest 10 messages
  * "see more"-action
  * spinning-wheel while it's loading
1. in the list-view: load matches/requests/conversations of a post only when it's folded-out
1. in the list-view:
  * only load the first 10
  * "see more matches/requests/conversations"-action
  * spinning-wheel while it's loading
1. in the grid/tiles view for matches:
  * only load the latest 10 matches *over all needs*
  * "see more"-action
  * spinning-wheel while it's loading
1. post-overview
  * only load 10 posts with the newest updates (where unread(?) messages trump unread(?) requests trump matches)
  * "See more posts"-action
  * spinning-wheel while it's loading
1. feed:
  * load 5 posts with the latest events – and 3 of the latter for each – with prioritization/sorting as specified above 
  * "see more events"-action for a given post
  * "see more posts"-action
  * spinning-wheels in all of the above places while they're loading

These intereactions should be introduced as new action-creators that load the data and then dispatch new actions of those new action-types. Some of these actions are triggered by clicks on the "more"-links, some by scrolling (e.g. in the feed)

The spinning wheels can be created by tagging the ownNeeds/post/connection-objects in the state with a `isFetchingMore` boolean.

m0ru:
Goal: have as few requests as possible.
 
Only load info that's relevant atm. E.g. skip closed connections / needs u

m0ru:
The bottle-neck atm: the crawling on the server: 
![unbenannt](https://cloud.githubusercontent.com/assets/1730492/16809833/0737e484-4922-11e6-9830-5bddb1ba5c3a.PNG)

counterparts and events that aren't included in the deep request:
![unbenannt](https://cloud.githubusercontent.com/assets/1730492/16809854/173a8ac6-4922-11e6-8c53-68ef7f04dba5.PNG)

m0ru:
Two different goals:
* As few requests as possible (which is important on low-bandwidth networks)
* minimal crawling effort on server (use the paging feature)

m0ru:
Searching for solutions:

[...]

PR #749, opened 2016-08-17
Paging in `won.fetch`

m0ru:
The first step for #623


quasarchimaere:
@m0ru can we delete this branch now, or should we leave it open since it is just a precursor for #623?

PR #750, opened 2016-08-18
Feat tags

### LINKED ISSUE #590: Show Description For Posts
Tasks:
- Add Post Description Field in Create View
- Show post description

### LINKED ISSUE #687: Show Tags for Post
show tags for post

### LINKED ISSUE #700: Use one triple per tag
Tags are currently stored in one triple as a comma-separated list. Should be one triple per tag

quasarchimaere:
fixes #700 fixes #687 fixes #590 

After creating a need with tags you should be able to see those tags within the post-info view (post-detail-view) of your need

Also the Description should be visible in the post-info now

Please also check if the Datastructure of won:hasTag (within the state) is an Array of Tags now instead of commaseparated string


webofneeds/won-owner-webapp/src/main/webapp/app/service/won-service.js
```diff
@@ -464,7 +464,7 @@ angular.module('won.owner').factory('wonService', function (
             title: need.title, //mandatory
             description: need.textDescription, //mandatory
             publishedContentUri: publishedContentUri, //mandatory
-            tags: need.tags.map(function(t) {return t.text}).join(','),
+            tags: need.tags.map(function(t) {return t.text}).join(','), //TODO: DO TAGS IN A DIFFERENT FORM
```
fkleedorfer: isn't this TODO resolved as of the PR #750 ?

PR #757, opened 2016-08-23
Geo distance query

### LINKED ISSUE #700: Use one triple per tag

### LINKED ISSUE #756: add location matching
* add location fields in solr index
* calculate distance in solr index between needs
* boost/filter matches according to distance

heikofriedrich:
added geo distance querying and date filering to solr query:
default query should work now like this:
* matching with different weights of content terms (title, description, tag fields)
* multiplicative (1 to 2) boosting of geo distance between needs
* date filtering: only match needs not more trhan 1 month apart
* location filtering: do not match needs with distance > 50 km
* can be tested by creating needs with different locations and checking the score the matches get (in the matcher service log file)
* added fix for tags in this pull request too => tag matching should work again, instead of stemming and synonyms the tag matching is now done with exact matching

fixes #756 
fixes #700


webofneeds/won-core/src/main/java/won/protocol/util/WonRdfUtils.java
```diff
@@ -744,6 +744,29 @@ public Boolean visit(Model model) {
 
       return ret;
     }
+
+    public static Float getLocationLatitude(Model need, URI needUri) {
+      Path propertyPath = PathParser.parse("won:hasContent/won:hasContentDescription/won:hasLocation/<s:geo>/<s:latitude>",
```
fkleedorfer: seems wasteful to parse this path on every call

PR #783, opened 2016-09-05
Feat#726 usedfortestingdebugbotchanges

### LINKED ISSUE #726: Allow flagging with `won:UsedForTesting` in GUI
We introduced the flag `won:UsedForTesting` in order to be able to separate test content from real content.

Currently there is no way in the GUI to add this flag.

Currently there are two ideas for what can be done:
# allow magic hashtag `#UsedForTesting`
pro: easy to implement, easy to learn
con: has to be entered every time, which is tedious if you are REALLY testing

# toggle 'test mode' in user settings. 
When test mode is `on`, all newly created needs get the `won:UsedForTesting` Flag
con: makes creating one single UFT need harder
con: harder to implement
pro: makes longer testing sessions easier

# toggle 'test mode' in GUI 
Actually, there is no need to go all the way to the user settings. It would suffice to have a 'test mode' toggle button (or something similar) somewhere where normal users don't really look (say, the bottom left corner of the page). Test mode should be associated with the session, and default to `off` in a new session.
pro: easy to implement
pro: makes creating many test needs easier
pro: makes creating only one test need only marginally harder
con: test mode has to be re-enabled upon login, which one may forget during testing

quasarchimaere:
added the debugbot changes in order to only create debugneeds when the UsedForTesting flag is set when creating a need (in owner-webapp won.debugmode = true)

fixes #726 (debugbot changes)

### DIFF
--- a/webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/debugbot/CreateDebugNeedWithFacetsAction.java
+++ b/webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/debugbot/CreateDebugNeedWithFacetsAction.java
@@ -44,66 +44,74 @@ import won.protocol.util.DefaultPrefixUtils;
 import won.protocol.util.NeedModelBuilder;
 import won.protocol.util.RdfUtils;
 import won.protocol.util.WonRdfUtils;
+import won.protocol.vocabulary.WON;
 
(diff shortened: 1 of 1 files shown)


quasarchimaere:
This is already pushed onto integration-test so it should be testable without further deployment

fkleedorfer:
tested successfully on integration-test

PR #794, opened 2016-09-08
Feat #791 mobilefirst changes

### LINKED ISSUE #791: MobileFirst Changes
Check/Review and Change the current GUI so that it works and looks nice on a mobile device as well

### LINKED ISSUE #795: Create Fixed Header

quasarchimaere:
fixes #791 
fixes #795 
this pull request is to heighten the gui support for mobile devices, this can be tested directly on your mobile device or with the use of most browsers (chrome, and firefox) when you set a specific device. another useful "tool" is the blisk browser (https://blisk.io/)

Be aware that this pull request also fixed the headers so make sure that none of the views we have is broken due to it before merging


heikofriedrich:
i tried it on the mobile phone, where it seems to look nice, but afterwards i tested in the normal browser and now i have a wrong display-mode there:

PR #799, opened 2016-09-09
Feat usewholeheight chat

quasarchimaere:
a chat window uses the whole height of the screen now and so you dont have to scroll two views


quasarchimaere:
chat

PR #802, opened 2016-09-09
Feat fix#733 postview jump

### LINKED ISSUE #733: Match @ post view  → jumps to overview in background
When you're in the detail view, go to the matches tab and then try to make a conversation request, the app will redirect you to the matches-overview before opening the pop-over. It shouldn't do that.

quasarchimaere:
Fixes #733 

please test all specific post tabs to see if the view stays at the specific post when viewing a connection or changing the layout type, before you always went to the overview page instead of staying at the specific page


m0ru:
Fails in post » matches » list view.

m0ru:
Fixed

PR #806, opened 2016-09-09
Feat selectiveevents2 #623

fkleedorfer:


fkleedorfer:
checked with @m0ru - works.

PR #867, opened 2016-10-19
Request reply need events

### LINKED ISSUE #860: Request/Reply mechnism of past need events for matching
* Solr matcher should be able to ask matching service at startup/restart what matches it has missed so far, these should be resend

heikofriedrich:
* fixes #860 
* can be tested: startup the whole environment normally, shutdown solr matcher, create needs and let matching service save or crawl them, then restart solr matcher, solr matcher should request these needs that were created while it was down. to check this check out its solr index for example
* tested this on integration test => seems to work


fkleedorfer:
Tried the steps, but in the solr matcher logs I'm finding this:
```
[INFO] [10/20/2016 21:31:46.702] [ClusterSystem-akka.actor.default-dispatcher-15] [akka://ClusterSystem/user/MatcherPubSubActor] initialized property 'lastSeenNeedDate' to -1, no need updates requested
```

heikofriedrich:
@fkleedorfer sorry the steps description was not accurate. This behavior is also intended, but i described the steps wrong. Before stopping the solr matcher at least one need has to be created so that it is sent to the solr matcher and updates its lastSeenNeedDate variable. Then it can be stopped and need can be created, and after restarting it should request and process them.

fkleedorfer:
so in the special case that the solr matcher sees nothing, this would happen:
1. solr macher is started, 
2. sees 0 needs because none are created
3. solr matcher is stopped
4. some needs are created
5. solr matcher comes up again
6. solr matcher does not ask for what it missed
If so, this does not sound right to me

heikofriedrich:
@fkleedorfer i made the change that if no need has been seen so far, the last need is requested to get the timestamp eventually 
=> pushed to integration-test

PR #870, opened 2016-10-19
Fix owner registration

fkleedorfer:
pushed that to integration-test, seems to work


fkleedorfer:
added a client side test to check if we need to re-register. The only remaining problem is that if the client cert changes, the client will re-register, which leads to stale data on the node (associated with the old cert), that will never be used again, yet the node does not know that. 

It would be ok for the node to just delete all data associated with owner applications after a timeout of, say, a month. However, I am not certain how the owner application would react to that.

fkleedorfer:
How to test:
* run owner+node, test with debug bot
* stop owner
* delete its owner-keys.jks file
* start owner
* check that it works with debug bot

You should find log messages (info) about the registration process. Look for `not yet registered`. You should find that during the owner's restart process

PR #894, opened 2016-11-18
Feat mailbot incl previousmessages

quasarchimaere:
feature to include last X textmessages of a connection in outgoing mails


quasarchimaere:
do not merge yet, there will still be minor changes (probably they will be completed next monday)

quasarchimaere:
can be merged now, no more additional changes pending

webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/mail/send/WonMimeMessageGenerator.java
```diff
@@ -18,16 +21,23 @@
...
 public class WonMimeMessageGenerator {
+    private static final Logger logger = LoggerFactory.getLogger(WonMimeMessageGenerator.class);
 
     @Autowired
     private VelocityEngine velocityEngine;
 
+    private int MAX_CONVERSATION_DEPTH = 3;
+    private char QUOTE_CHAR = '>';
+
+    private static final String YOUR_MESSAGE = "You said: ";
```
heikofriedrich: this static variables should go into the templates right?

heikofriedrich: i mean this "You said" "They said"

quasarchimaere: i think this can be left alone for now, i do not think we need to transfer those vars to the properties or the templates since we will have to think about integrating i18n at some point and thus will have to change the way we include "static" text anyway

PR #902, opened 2016-11-23
User mail subscription

### LINKED ISSUE #882: save status of user responses in bot context
* if user unsubscribed, mail bot should not send him messages again
* if user responded with "yes publish my need", need and following needs should be published
* until user answers mails have to be saved temporerly or let on the mail server marked somehow so that they can be easily retrieved and published when the user answers

### LINKED ISSUE #889: unsubscribe footer for bot mails
* change template
* save info to which user no mails should be send 
* before sending mail check list

heikofriedrich:
* fixes #882, fixes #889 

* added functionality for the user to subscribe/unsubscribe to the mail bot, meaning publish all my future needs / ignore mail bot in future (opt-in/opt-out)
* created first draft of welcome mail that is sent in response to mails on the mailinglist if user has not subscribed or unsubscribed yet 
* added footer to hint/connect/message mails to send and close event using mailto links (close need functionality is not yet implemented)


quasarchimaere:
as far as i saw the opt-in/opt-out strategy is not adaptable for now (always explicit opt-in first) maybe we can configure this with another pull-request later

PR #906, opened 2016-12-01
improved cache headers and landingpage config in owner application

fkleedorfer:
improved caching headers in owner application to allow for client side caching of static resources

How to test: open the owner application, inspect the network information in the browser. After the first page view, almost all resources should cause a 304 not modified response upon page reload. 
If the page is requested again (via entering the url again), most resources should be served from the cache directly.


m0ru:
chrome seems to ignore the header

PR #907, opened 2016-12-01
Feat better linkeddatacaching

fkleedorfer:
enabling client-side caching for linked data requests

For now only implemented for Event and Connection. 
Requires adding a version field to the connection and the rdf_datasets table in the node database.


quasarchimaere:
will be merged once we introduce flyway database migration

quasarchimaere:
merge #929 and #930 before this one (#930 is the branch that the comment before refers too, #929 is the telegram bot (to see fewer changes for #930)

webofneeds/won-core/src/main/java/won/protocol/model/Connection.java
```diff
@@ -33,6 +33,10 @@
   @Column( name = "id" )
   private Long id;
 
+  @Version
```
heikofriedrich: changes that affect the database must now (after integration of flyway pull request) also provide a new flyway migration script => this must be added if this pull request is integrated after flyway pull request

fkleedorfer: yes I agree, let's keep this PR open until we have that

fkleedorfer:
Continued work in #1021

PR #911, opened 2016-12-14
Fixed bug in selector used by "show more".

m0ru:
Apparently for some success-responses the corresponding-remote-message is only saved as string. This is a quick fix, that allows a selector used by the "show more"-button to deal with these without crashing.

Test: open connection, chat around a bit, click show-more -> no more exceptions. Make sure to test with the master-branch first, so you have a events on your connections that cause the exception to begin with.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/selectors.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/selectors.js
@@ -8,6 +8,7 @@ import won from './won-es6';
 import {
     decodeUriComponentProperly,
     msStringToDate,
+    is,
 } from './utils';
 
 import {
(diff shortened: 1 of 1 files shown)


fkleedorfer:
It's unclear to me
* in which way the bug manifests - how do I know it happened?
* how I reproduce the bug?
May that be a server-side problem we should look into more closely?

m0ru:
Hm, can't reproduce it any more either. :|

fkleedorfer:
so what do we do? If the commits constitute an improvement, we merge them (after testing) if not, we just close the PR. The change looks useful to me. @m0ru can you comment?

heikofriedrich:
i can reproduce it (or is it a similar bug on master). In all my needs when i click in a conversation on show more a get an javascript exception:

angular.js:5209 TypeError: Cannot read property 'filter' of null
    at connections-actions.js:298
    at index.js:13
    at router-middleware.js:43
    at Controller.connections__showMoreMessages (bindActionCreators.js:8)
    at fn (eval at compile (app_jspm.bundle.js:1), <anonymous>:4:445)
    at callback (angular.js:9694)
    at Scope.$eval (angular.js:7617)
    at Scope.$apply (angular.js:7640)
    at HTMLAnchorElement.<anonymous> (angular.js:9699)
    at HTMLAnchorElement.dispatch (jquery.10.2.js:5109)

heikofriedrich:
showmore

heikofriedrich:
created similar bug #958

m0ru:
Yeah, I think we should merge.

m0ru:
Contains a typo, that's fixed with #973

PR #965, opened 2017-02-02
fixed defect #945

### LINKED ISSUE #937: Fold in "Sign In" when clicking on "no account yet? sign up"
steps to reproduce:

1. go to landing page
1. click on sign in
1. at the bottom of the sign-in-mask click on "no account yet? sign up"

result should look something like this:

glitch

### LINKED ISSUE #945: login popover goes away when browser inserts wrong password
To Reproduce: Make the browser remember a wrong password and then have it fill it in. The login popover disappears and so it looks as if the password had been correct and the page is loading. But actually if you open the popover again, you see the error message.

quasarchimaere:
fixes WRONG#"945"WRONG, hides the login"drawer" when pressing signup within the login
fixes #937

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/topnav.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/topnav.js
@@ -59,7 +59,7 @@ function genTopnavConf() {
                         <li ng-show="!self.loggedIn">
                             <a class="topnav__button"
                                 ng-click="self.showLogin()"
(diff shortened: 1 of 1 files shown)


heikofriedrich:
deployed it on integration-test => still doesnt work for me. after a wrong password is entered the popup is closed and nothing happens ...

quasarchimaere:
there was an mistake on my end, the fix does not resolve bug #945 but fixes bug #937, i am sorry for the confusion

PR #982, opened 2017-02-08
Fix - Layers.png

### LINKED ISSUE #961: layers.png not found

m0ru:
Added gulp-task that copies imgs needed by osm.

Resolves #961


webofneeds/won-owner-webapp/src/main/webapp/gulpfile.js
```diff
@@ -54,6 +54,18 @@ gulp.task('sass', function(done) {
...
 });
 
+/**
+ * Copies over static resources, that libraries in
+ * the bundle need to be in a place relativ to
+ * themselves -- and thus the bundle.
+ */
+gulp.task('copy-static-res', function(done) {
+    return gulp.src([
+        './jspm_packages/npm/leaflet@0.7.7/dist/images/**/*'
```
heikofriedrich: is getting this resource in the right place not possible without a static reference to the version number of leaflet? we should avoid this since the error might reoccur when the leaflet version is updated without updating this code

m0ru:
The task doesn't depend on the concrete leaflet-version any more, but rather extracts it from systemjs.

PR #996, opened 2017-02-17
Feat mail origin check

quasarchimaere:
implemented a check that validates a sent command mail from adress with the original message that the need has been created with

removed a verbose system.out logprint of a dataset


fkleedorfer:
closing as it is part of #1000

PR #1000, opened 2017-02-20
Feat close need on taken

quasarchimaere:
implemented the possibility to close an open need via an e-mail that matches a takenCmdPattern (currently set to [TAKEN]) in the subject line, the mailbot will then crawl through all the created needs of that e-mail adress and close the first open need that matches the title e.g "[WANT] testneed" will be closed when the same e-mail adress sends a "[TAKEN] testneed" e-mail

### DIFF
--- a/webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/mail/receive/MailCommandAction.java
+++ b/webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/mail/receive/MailCommandAction.java
@@ -15,14 +15,18 @@ import won.bot.framework.eventbot.event.impl.mail.CloseConnectionEvent;
 import won.bot.framework.eventbot.event.impl.mail.MailCommandEvent;
 import won.bot.framework.eventbot.event.impl.mail.OpenConnectionEvent;
 import won.bot.framework.eventbot.event.impl.mail.SubscribeUnsubscribeEvent;
+import won.bot.framework.eventbot.event.impl.needlifecycle.NeedDeactivatedEvent;
 import won.protocol.util.WonRdfUtils;
 
+import javax.mail.Address;
 import javax.mail.MessagingException;
+import javax.mail.internet.InternetAddress;
 import javax.mail.internet.MimeMessage;
 import javax.naming.AuthenticationException;
 import java.io.IOException;
 import java.net.URI;
 import java.security.AccessControlException;
+import java.util.List;
 
 import static jdk.nashorn.internal.runtime.regexp.joni.Config.log;
 
@@ -67,7 +71,12 @@ public class MailCommandAction extends BaseEventBotAction {
             case UNSUBSCRIBE:
                 bus.publish(new SubscribeUnsubscribeEvent(message, SubscribeStatus.UNSUBSCRIBED));
                 break;
-
+            case CLOSE_NEED:
+                URI needUri = retrieveNeedUriFromMail(message);
+                if(needUri != null) {
+                    bus.publish(new NeedDeactivatedEvent(needUri));
+                }
+                break;
             case NO_ACTION:
             default:
                 //INVALID COMMAND
@@ -108,7 +117,10 @@ public class MailCommandAction extends BaseEventBotAction {
             logger.debug("Command Message Sender: "+message.getFrom());
             logger.debug("Original Message Sender: "+originalMessage.getFrom());
 
-            if(!message.getFrom()[0].equals(originalMessage.getFrom()[0])) {
+            String senderNew = ((InternetAddress) message.getFrom()[0]).getAddress();
+            String senderOriginal = ((InternetAddress) originalMessage.getFrom()[0]).getAddress();
+
+            if(!senderNew.equals(senderOriginal)) {
                 throw new AccessControlException("Sender of original and command mail are not equal");
             }else{
                 logger.debug("Sender of original and command mail are not equal, continue with command processing");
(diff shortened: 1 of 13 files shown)


webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/mail/receive/MailCommandAction.java
```diff
@@ -91,22 +134,56 @@ private void processReferenceMailCommands(MimeMessage message, String referenceI
...
+        }
+    }
+
+    /**
+     * This Method tries to find a corresponding open need uri from a user(given by the from adress) and returns the
+     * corresponding need uri if there was an open need with the same title
+     * @param message used to extract sender adress and subject(title)
+     * @return
+     */
+    private URI retrieveCorrespondingNeedUriFromMailByTitle(MimeMessage message){
```
heikofriedrich: im not sure if we should implement the taken/close use case by using the same title. what happens if there are more needs with the same title? also what happens if the message title does not exactly match the need title? 
=> i have already started to write code to take care of this problems using message ids and mailto links that the user should click on that then produce answer mails with specific subjects that hold an message id. 

E.g. when you look at the footer of some mail (e.g. match found mail) there is this:

---
close this request and do not receive mails for it anymore:<mailto:mail2won.int@gmail.com?subject=close_Message-Id_%3C1120133180.3.1487771411744.JavaMail.root%40974cfab3b863%3E>
do not publish my requests on this mailinglist and ignore mails from Web of Needs Integration Test Mail Bot in the future: <mailto:mail2won.int@gmail.com?subject=unsubscribe>

first link is for example for closing and refers the original mail by using a message id.
im not so sure how we combine this concept with closing by title ...

fkleedorfer: given how mailing list users currently do it in many cases, this feature ist simply required. I mean, it is common practice to announce the availability of something using an email subject and later announce it's no longer available: 
```
OFFER: Children's clothes
...
TAKEN: Children's clothes
``` 

Using the TAKEN (GONE/VERSCHENKT/...) subject as a trigger for our 'deactivate' message seems safe to me. If a user sent multiple mails with the same subject, thereby creating multiple needs with the same title, the TAKEN mail will close the first one.

fkleedorfer: ... and if the subject lines aren't exactly equal, it may be ok to send an email to the user explaining that we wanted to deactivate a need but did not know which one, and she may want to review her needs/mails (but for now just not deactivating anything is ok too)

heikofriedrich: ok this seems fair enough too me

heikofriedrich:
i deployed and tested the pull request on integration test and locally.
i created a need via mail. checked the rdf on the won node (state active) and afterwards send a TAKEN message to the mail bot. The mail bot closes the need => need rdf on won node switches to inactive. 
The problem is then the mail bot enters some kind of loop in which it tries to register itself continuesly at the won node and spams the log. This should be fixed. Output is something like:

22.2.17 14:33:09.008 [taskScheduler-10] INFO  w.o.m.OwnerProtocolCommunicationServiceImpl - setting up communication with won node: https://satsrv04.researchstudio.at:8889/won/resource
22.2.17 14:33:09.009 [taskScheduler-10] INFO  w.o.m.OwnerProtocolCommunicationServiceImpl - we're already registered. Connecting with WoN node: https://satsrv04.researchstudio.at:8889/won/resource
22.2.17 14:33:09.017 [taskScheduler-10] INFO  w.o.m.OwnerProtocolCommunicationServiceImpl - connected with WoN node: : https://satsrv04.researchstudio.at:8889/won/resource
22.2.17 14:33:09.026 [taskScheduler-3] INFO  w.o.m.OwnerProtocolCommunicationServiceImpl - setting up communication with won node: https://satsrv04.researchstudio.at:8889/won/resource
22.2.17 14:33:09.037 [taskScheduler-3] INFO  w.o.m.OwnerProtocolCommunicationServiceImpl - we're already registered. Connecting with WoN node: https://satsrv04.researchstudio.at:8889/won/resource
22.2.17 14:33:09.046 [taskScheduler-3] INFO  w.o.m.OwnerProtocolCommunicationServiceImpl - connected with WoN node: : https://satsrv04.researchstudio.at:8889/won/resource
....

quasarchimaere:
The loop happened because we where calling and listening on the wrong event, namely the deactivatED need event, that called another needclose event, thats why it looped

heikofriedrich:
i dont know what the getMailReference method returns but i guess if there is not a reference number there the problem could be fixed if null is returned then.

quasarchimaere:
i fixed the issue @heikofriedrich was referring too, this should work now, but please make sure normal communication (for example with a debug need, still works fine, as the parts that have been changed also possibly influence these commands)

heikofriedrich:
i fixed a small null pointer bug but i found another problem. Here is how to reproduce:

Start wonnode, mail bot and debug bot up. Create a mail and send to the mail bot. Wait for confirmation mail and subscribe (click on subscribe link and send mail back) so that future needs are published. Then create a debug need so that you get a match or conversation request mail back. Click on the mailto link in the footer "close this request and do not receive mails for it anymore" and send mail back that you want to close the need. This action doesnt close the need but it should. This is another closing option instead of TAKEN in the subject => is this one still needed besides TAKEN? Do we need this mail identification by message-ids or can we just replace it by only one way of closing a need?

PR #1021, opened 2017-03-20
Feat improve db and linkeddata handling

fkleedorfer:
This comprises everything from #907 

This PR contains db changes. If testing on an existing db, check if the schema update worked.
Otherwise, queries to linked data entities should be cachable and updatable based on etags.


fkleedorfer:
pushed that on integration-test btw. Seems to work for me

quasarchimaere:
commit of jspm config is obsolete as it is altered while building anyway, merging pull request now

PR #1046, opened 2017-04-10
improved normalization

fkleedorfer:
now handles different types of cycles and references across branches correctly
There is still an issue, however, with orphaned rdf nodes after normalization. In the current
use case (the nodes reachable from the need node are extracted with graphExtract) this is not
a problem


heikofriedrich:
tested if the normalization tests work and if the normalization works for my sample needs => merged

PR #1060, opened 2017-04-13
Need structure impl gui

m0ru:
Theoretically the refactoring is complete, but I've either created or stumbled over a few bugs while doing it (dunno how we want to procede with these):

* When a connection goes from "Suggested" to "RequestSent" it's currently not correctly updated in the connection-reducer. A page-reload is necessary for it to show up in the "sent-requests" tab.
* The chat-textfield loses focus after sending a message. 
* The chat does not fully/correctly scroll down. There might be some problem with how the `$watch` interacts with the state-updates.
* The grid/flow-layouts have graphical glitches on chrome (the letter of the image wraps into the next column) and only layouts as a single column in FF. We could try the css-grid-layout here, that should be [supported by almost all browsers](http://caniuse.com/#feat=css-grid) by now.
* There's some double borders, i.e. where the neighbouring component also has a border, thus resulting in a 2px-wide border that looks unappealing and that visually pops out. This is the case between e.g. right-side and chat.
* When the list on the left-side is longer than the chat, the lack of a border on the right edge of selected items (intended to deal with the problem above) becomes apparent. 
* The context-menu that's now available on all connection-list-items, should be relabeled depending on the context (e.g. for matches it should read "remove")
* sometimes need-titles still end in `&nbsp;` or `<br>`, due to problems with mediumjs / contenteditable.

Also there's a few refactoring opportunities that are still open:

* the right side of incoming-request, sent request and chat are similiar in that they have (almost) the same title and an "x"-button to close them. This could be directive that allows custom html inside.
* matches-flow-item and matches-grid-item are almost the same from a functional viewpoint and could be merged to a single component. The difference is only the size of the title-picture (and presence of a gallery?). 
* connection-selection is really only a special case of connection-overview. The latter would suffice, if it hid the own needs and removed the intendation of counterparts, when used on a detail page.


fkleedorfer:
Bug in Feed - firefox (not chrome): 
Create a need in debugmode. Wait for the bot to connect. Reload the page. One of the two connections (hint or connect) is not loaded correctly, leading to a black '?' and no title to be shown in the feed view.

The JS Console contains this error message, which always mentions the connection that could not be loaded:
```
"Couldn't resolve https://192.168.124.49:8443/won/resource/connection/gjuzeokq5cfpz1xf3fqa. reason: {"msg":"Failed to fetch https://192.168.124.49:8443/won/resource/connection/gjuzeokq5cfpz1xf3fqa","causedBy":"failed to load https://192.168.124.49:8443/won/resource/connection/gjuzeokq5cfpz1xf3fqa due to reason: \"SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data\""}"
```
When fetching that URL manually, all seems fine.

PR #1068, opened 2017-05-10
Bot context wrapper

quasarchimaere:
Changed bots to use a BotContextWrapper instead of the plain botcontext bean, and moved all methods from eventbotactionutils as well as defined list names to the respective botcontextwrappers -> these are now also configurable over spring config xml and not only via constants in the sources


quasarchimaere:
i will change the botcontextwrapper config and implementation so one only has to define a botname and not every individual list name, as well as an updated bot/readme.md, i will close this pull request for now

Issue #1052, opened 2017-04-13
Title ends with `<br>`


fkleedorfer:
Not every title, but almost every title I create in the GUI is somehow transformed to have trailing `<br>`

m0ru:
One way to reproduce this, is to type "asdf #asdf" (without ending in a linebreak). The tag will will then contain a "<br>"

fkleedorfer:
or to use firefox

m0ru:
ah, yes, the problem is FF specific.

Issue #1042, opened 2017-04-03
contenteditable/medium.js: sometimes not all lines are `<p>`s


m0ru:
sometimes in chrome (when typing fast?) i get the following
```
<div class="medium-mount Medium Medium-partial" contenteditable="true">
  <p class="medium_title">&nbsp;asdf #asdf</p>
  qwerqwer #qwer qwerqwer
</div>
```
Which causes the parsing of the description and it's tags to fail (silently), as the selector `'.medium-mount p'` doesn't match the last line that isn't wrapped in a `<p>`-tag.

m0ru:
can't reproduce it atm :|

quasarchimaere:
no longer relevant as we removed medium.js from our codebase

PR #1086, opened 2017-05-15
Using textContent instead of innerHTML to get title.

### LINKED ISSUE #1080: need-titles ending in  `&nbsp;` / `<br>` - contenteditable strikes again
sometimes need-titles still end in `&nbsp;` or `<br>`, due to problems with mediumjs / contenteditable.

Related to #1042

m0ru:
Test: go to create-need in FF, enter title like "asdf #asdf" and don't line-break. before this fix, the tag would contain a "\<br\>". Alternatively you can just create a need with a title and check how it turns out in any post-specific view where the title shows up.

Resolves #1080, #1052


m0ru:
Resolves #1042
Resolves #1052

Issue #1091, opened 2017-05-15
Text on incoming request only shown after reload


m0ru:
1. create need with debug-flag
2. accept connection
3. in messages tab, the "Greetings! I am the DebugBot."-message shouldn't be there (even when you click "show more")
4. reload page
5. it should be displayed now

m0ru:
same goes for the incoming-request view, btw (there should be two paragraphs: the description and the request message)

apparently that event isn't in the state, without reloading, as you can check by pasting the following snippet while in the incoming request page:

```javascript
var state = pc4dbg.$ngRedux.getState()
var hasCnct = e => e && e.get && (e.get('hasSender') === pc4dbg.connectionUri || e.get('hasReceiver') === pc4dbg.connectionUri)
state.getIn(['events','events']).filter(e => 
  hasCnct(e) || hasCnct(e.get('hasCorrespondingRemoteMessage'))
).toJS()
```

m0ru:
The necessary data arrives in the `onMessage`
```
{  
   "msg:FromOwner":{  
      "hasContent":"https://192.168.124.53:8443/won/resource/event/575cvhkp9ywa1ypq5g9d#content-lwc8",
      "hasMessageType":"msg:ConnectMessage",
      "hasReceivedTimestamp":"1495120121686",
      "hasReceiverNeed":"https://192.168.124.53:8443/won/resource/need/6800507546094662000",
      "hasReceiverNode":"https://192.168.124.53:8443/won/resource",
      "hasSenderNeed":"https://192.168.124.53:8443/won/resource/need/2lti4jaq7qw5mup10xcy",
      "hasSenderNode":"https://192.168.124.53:8443/won/resource",
      "hasSentTimestamp":"1495120121621",
      "protocolVersion":"1.0",
      "hasFacet":"won:OwnerFacet",
      "hasRemoteFacet":"won:OwnerFacet",
      "hasTextMessage":"Greetings! \nI am the DebugBot. I can simulate multiple other users so you can test things. I understand a few commands. \nTo see which ones, type \n\n'usage'\n\n (without the quotes).",
      "uri":"https://192.168.124.53:8443/won/resource/event/575cvhkp9ywa1ypq5g9d",
      "framedMessage":{  
         "@context":{  },
         "@id":"https://192.168.124.53:8443/won/resource/event/575cvhkp9ywa1ypq5g9d",
         "@type":"msg:FromOwner",
         "msg:hasContent":{  },
[...]

PR #1100, opened 2017-05-22
Fix connect message #1091

m0ru:
The message bot's connect message "Greetings! I am the DebugBot..." should now be visible in open chats and the incoming request tab *without* reloading the page (see #1091 for further details).


m0ru:
Resolves #1091

Issue #1074, opened 2017-05-15
chat-textfield loses focus after sending a message


m0ru:
This makes it hard to send multiple messages / needlessly requires moving the hand to the mouse after every one.

PR #1103, opened 2017-05-26
Refocusing chat-txt-field after sending a message.

m0ru:
When writing chat messages, the cursor should refocus the text-field, thus allowing continous writing without moving the hand to the mouse.


m0ru:
Resolves #1074

m0ru:
Resolves #1074

PR #1124, opened 2017-06-09
Fix message references

fkleedorfer:
improved and documented the message referencing rules


fkleedorfer:
@m0ru this will help your work on message ordering in the owner webapp

m0ru:
Ran into an exception when trying to deploy: https://gist.github.com/m0ru/d8a3744de14110c912042002a6202cd4 

I reimported the maven config and rebuild everything before. Are there any config-changes necessary?

fkleedorfer:
I don't think so, but a mvn clean is probably a good idea.

m0ru:
Yep, that fixed it ^^"

PR #1128, opened 2017-06-12
Making sure we sort by timestamps in own(ed) events

### LINKED ISSUE #402: front-end: displayed order of text messages sometimes wrong

m0ru:
Mostly code styling/refactoring. The relevant fix is 3361722, that makes sure we sort based on the timestamps of our own owner/server (i.e. the "outer" events, not the ones that are nested as `hasCorrespondingRemoteMessage`)

Also contains a fix, for the "pending" status -- it wasn't displayed before. Some of the out-of-order messages might be pending ones, as these are sorted to the bottom of the list. We might want change this behaviour and instead show them where their `sentTimestamp` would place them. In any way, we now can see when this happens for that reason.

Resolves #402


m0ru:
@sorting by "pending"-status: the idea was to prevent jumping in the message order, as messages that are confirmed delayedly will have that later timestamp and jump to the bottom. However, it also causes the problem, that unconfirmed messages can "pile up" at the bottom, thus occluding the chat. I've removed this sorting criterium. We can reintroduce it, once we have some sort of timeout for the success-responses, so messages can't stay "pending" forever.

PR #1145, opened 2017-07-04
removed unused state for open/close posts -> will be at component lev…

quasarchimaere:
remove the open/close posts from state -> up until now we saved the overview post list visibility in the state, but never used it, i removed the unused reducer, the view (post-overview) should still have two collapsible panels (for open and closed posts)


fkleedorfer:
tried the code on integration-test, seems to work.

PR #1162, opened 2017-07-17
Refac state remove connections

quasarchimaere:
removed connections from state (connections are now only retrieved from the ones within needs)


m0ru:
notification count -- red 0
^ we should either remove the color highlight or -- preferably -- not show the number when it's 0.

sign-in: when entering a wrong password, the popup still closes, thus hiding the error message

1 hint but notification counter at 0
^ notification counter for hint's failed

bot's request's message not shown
^ this should read: "This is a need automatically created by the DebugBot." (description) and "
Greetings! I am the DebugBot...." (message)

{wip}

m0ru:
Imo not that many regression bugs, given a refactoring of this pervasiveness and the low refactoring support javascript allows.

quasarchimaere:
issue 2 is a bug that was present before refactoring as well (issue #945)
issue 4 and issue 6: are having the same ( we need to search through the events and store the connect messages within need-reducer, after storing the connection)
issue 8: is not reproducible for me (no line of code found within our code that uses length instead of size)
issue 9: is probably just a lag/data retrieval issue

please recheck the pull request and test the fixed issues,and make bug entries for the still existing issues, as they are not really blocking in my opinion

PR #1173, opened 2017-07-21
Feat accountless posting

m0ru:
Create Need is now visible at all times. When you post without being signed in, the app creates an account with random username and password. You should be able to chat with the debug bot this way.

Not implemented:

* [ ] signing in with Id on page reload
* [ ] putting create-need-widget onto the landing-page / replacing landing-page
* [ ] upgrade to full account


m0ru:
Need to get page-reload working.

m0ru:
Working page-reload.

fkleedorfer:
Cannot connect with a match - the 'good - request conversation' button does not seem to do anything

fkleedorfer:
When I create a need in accountless mode and then close it, I get matches to it when I create another - matching - need

fkleedorfer:
The message ordering is not correct, not even after reloading. We are reading the wrong timestamps, it seems.

fkleedorfer:
In the feed, when I click on a matched need, the view changes to the matches view (good) and the overlay that allows me to connect with that need is shown (good). I can then connect with the need (good), but the overlay stays. And I think the counters are not updated.

fkleedorfer:
In the incoming request view, I cannot decline an incoming request (click the decline button, nothing happens)

fkleedorfer:
all of the above happen also in normal account mode

m0ru:
TODOs

* [x] request conversation
* [ ] match with own closed
* [ ] message ordering
* [ ] match-overlay staying
* [ ] declining incoming request

m0ru:
Most of these were me mechanically translating `ui-srefs` into `ng-click="self.stateGo...(...)"` without checking if there's alreay another `ng-click` on the element.

m0ru:
counters, message ordering and matches with own closed needs shouldn't be something that this pull-request has touched and probably are general regression bugs. can you open tickets for these?

PR #1193, opened 2017-08-31
Updated angular version

m0ru:
I've **updated the version of angular** (and any libs depending on it). Apparently, we have multiple version of several libraries installed (`jspm install` prints them as **"Installed Forks"**) -- I've changed `package.json` so at least we don't have multiple ones of actively installed libraries.

The update was necessary to be able to use multiple ng-transclude (e.g. `<dropdown><dd-header>{some html}</dd-header><dd-menu>{some html}</dd-menu>`) required for the refactoring/unification of the login/logout-menus => **Please leave the branch open** after merging :)
(I thought I'd rather open the pull-request earlier than later, so the version changes don't cause more problems than necessary)

NOTE: due to the change to redux-ui-router the router-object in the state isn't immutable any more, i.e. `state.getIn(['router',...])` won't work any more. I've changed the **`getIn`** in `utils.js` to be able to handle objects with mixed (im)mutability. Preferably use this function wherever possible, so we don't have to deal with similar issues in the future :)


m0ru:
Opened a pull-request (#1194) with a distinct branch that can be deleted afterwards.

PR #1194, opened 2017-08-31
Refac angular update

m0ru:
I've **updated the version of angular** (and any libs depending on it). Apparently, we have multiple version of several libraries installed (`jspm install` prints them as **"Installed Forks"**) -- I've changed `package.json` so at least we don't have multiple ones of actively installed libraries.

The update was necessary to be able to use multiple ng-transclude (e.g. `<dropdown><dd-header>{some html}</dd-header><dd-menu>{some html}</dd-menu>`) required for the refactoring/unification of the login/logout-menus. :)

NOTE: due to the change to redux-ui-router the router-object in the state isn't immutable any more, i.e. `state.getIn(['router',...])` won't work any more. I've changed the **`getIn`** in `utils.js` to be able to handle objects with mixed (im)mutability. Preferably use this function wherever possible, so we don't have to deal with similar issues in the future :)


quasarchimaere:
deployed on integration-test, leaflet does not seem to set the marker anymore, is this a known issue?
clicking on a need from the feed also results in an empty post-info page (need loading doesnt work)

quasarchimaere:
Remark: is the unification/refactoring of the login/logout dropdowns even in the scope of the current uf-milestones?

fkleedorfer:
@quasarchimaere yep we wanted to put the about link into the menu, but the menu is fighting back, so we need to show it who's the boss for a little bit

m0ru:
@link in feed: apparently we'd lucked into the only version of ui-router before, where interpolating a string in an sref worked accidentally. There's been some discussion over at [ui-router#395](https://github.com/angular-ui/ui-router/issues/395). Hower the approaches using `ui-sref` and `ui-state` don't seem to work (well) for our use-case. I've added utility functions that directly make use of `$state.href` to generate hrefs (and those we can interpolate all day long :3 )

Apparently, also ui-router's stateChange-events where removed [in favor of a transition-API](https://ui-router.github.io/guide/ng1/migrate-to-1_0#state-change-events) (which mainly impacts the hook for our access-control)

Issue #546, opened 2016-05-18
Page-load performance optimisation


m0ru:
Here's a few enhancements we could make to drastically increase the page-load duration, roughly sorted by how much effort they are to implement:
1. Minify the bundle (expecting 50%-70% saved in size)
2. Make sure server gzips the bundle
3. Make sure the big bundle is cached by the browser (saves time on reloads)
4. Everything loaded by the load-action as pre-baked script exporting an initialisation object, included directly in index.jsp. This saves us a server round-trip.
5. Find out which scripts are bloating the bundle's size and replace them with slimmer alternatives.
6. divide the bundle into chunks.first load scripts necessary for the actively displayed view, then load rest in a second pass while the user is still on that initial view.

fkleedorfer:
This is really a showstopper for many people.

m0ru:
An first step that we should take before starting with service-workers, would be to generate a hash of every built `bundle.js` and append that to it's name and statically generate `index.html` with `<script src="bundle-a7asd875asdf.js"></script>`. Then we can set the caching-policy with which it's served to `"forever"`.

m0ru:
We could add the [clojure-compiler](https://developers.google.com/closure/compiler/) to the build. Among other things it does limited tree shaking and minification. For a different project I'm currently working on, this reduced the app-bundle size from 2.2MB to 550kB (140kB after gzipping). Dunno if we can't also get babel/jspm to minify/tree-shake.

m0ru:
Also, we should go over our javascript dependencies, check their file-size and try to replace the larger ones with slimmer alternatives.

m0ru:
Related issues:

* Optimize landing page #987
* Optimise images to speed up page-load #978
* Tidy out dependencies in owner-webapp #1047

m0ru:
As for minification: there's a jspm-option for that. Just add `minify: true` to the call to `gulp_jspm(..)` in `gulpfile.js`. This will cause some compilation problems though, that need to be fixed (e.g. missing semicolons).

m0ru:
jspm allows us to split bundle. If angular allows adding components later at run-time, we could pull the landing page it's dependencies into a seperate bundle and load the rest of the app in a second pass. Dunno how we'd react to people entering via the create-need or post-detail pages? have a initial-deferred pair for every possible entry-point?

fkleedorfer:
The worst part is loading all the data. Possible approaches:
[...]

PR #1203, opened 2017-09-08
Feat unified js

m0ru:
Medium.js is capable of being bundled (without extra config) now! This allows us to load bundle directly after the html finishes loading/parsing.

**before:**

loading_landingpage_-_good_3g_-_with_gzip

(see [here](https://github.com/researchstudio-sat/webofneeds/issues/546#issuecomment-327556409) for explanation of the experiment))

**after:**

loading_landingpage_-_good_3g_-_with_gzip_and_bundled_mediumjs

The **1 second spent queuing** for the bundle-download is gone. For some reason the bundle download is faster now but the sign-in request takes a lot longer.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/chat-textfield.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/chat-textfield.js
@@ -9,6 +9,7 @@ import angular from 'angular';
 import 'ng-redux';
 import Immutable from 'immutable';
 import 'angular-sanitize';
+import Medium from 'medium.js';
 import {
     dispatchEvent,
     attach,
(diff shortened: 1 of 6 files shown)


quasarchimaere:
Can we do some investigation on why the sign in process takes a lot longer now? i think roughly 8 sec for the check isSignedIn is not that great. i would hold off on merging this pull request until further investigation/changes have been made

m0ru:
Hm, it's the second `checkLoginStatus` triggered in `load-action.js:60` (only shortly after the one in `configRouting.js:220`). This is repeatable, even if I remove one or the other request. Theoreticall I could make the function stateful to cache the result until a sign-in/sign-up/sign-out has been done, tough I'd rather avoid that -- maybe there's a better way to avoid the redundant request. My approach to diagnosing would be to print timestamps in `RestUserController.isSignedIn` and checking where the time is lost as the contrast between the two requests is rather stark.

m0ru:
Nvm, it's just chrome-devtools showing bogus. If I log `Date.now() - startTime` when the check succeeds or fails, it takes 670ms.

m0ru:
#1206 gets rid of the redundant request (the first one though, i.e. not the one for which Chrome's measurement breaks ^^")

PR #1236, opened 2017-10-03
Revert "Request Contact Without Need + Faster SPARQL Query"

quasarchimaere:
Reverts researchstudio-sat/webofneeds#1235


quasarchimaere:
no revert needed

PR #1238, opened 2017-10-04
Fix #1220 whatsaround disabled tracking

### LINKED ISSUE #1220: whatsaround needs not working if location tracking is disabled
browser like firefox asks if you allow location tracking if you first enter the website matchat.org. if you answer "no" and create whatsaround need, then the need creation is hanging in the gui "Retrieving Whats Around ..." but does nothing, it could be confusing for the user

quasarchimaere:
fixes issue #1220 
fixes location need-builder typo


m0ru:
That's good :)

Error messages should advise on a recovery strategy (reloading and accepting, or checking the browser preferences in this case). I've added a commit that extends the message :)

In general, we could also offer the location picker here, if location access isn't allowed.

PR #1252, opened 2017-10-11
Fix reply message #1241

m0ru:
All messages exchanged in the connect/open back-and-forth should be shown in the chat to both sides.

Resolves #1241

PR #1253, opened 2017-10-11
Fix  relative time stamps

m0ru:
The chat-messages should have relative timestamps (e.g. "just now", "5 minutes ago") again. Timestamps should (also) show again for matches, the post-info-page, etc.

NOTE: merge after #1252


quasarchimaere:
the relative timestamps would work but since this contains the pull request #1252 i can't merge this yet

PR #1259, opened 2017-10-16
Feat create post update

### LINKED ISSUE #922: Location-picker should be shown optionally
See Ulf's prototype. The location picker, along with all other optional fields, should only be visible if the user asks it to be.

### LINKED ISSUE #1183: Create Need - Styling optional-fields area
Atm the the location-picker isn't styled (results should be a drop-down, `[CHECK]` should be a check-mark instead) and the optional fields aren't added in the way layed out in the paper prototype. 

Should we start using the intended design variant, we can add the other options as greyed out buttons to suggest future features and communicate the intention of the area. On the downside, having unusable interaction elements might be frustrating, though. Then again, they could inspire people to add that information to the free-form description field.

quasarchimaere:
adds animations and more detail to the create-view
adds a new field for tag input (for users who dont know how to add tags via the description)
ability to remove details in order to remove the location/tags from the post (if the user chooses not to include them)

fixes #922 
fixes #1183

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-post.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-post.js
@@ -4,6 +4,8 @@
 ;
 
 import angular from 'angular'
+import ngAnimate from 'angular-animate';
+
 import 'ng-redux';
 import createNeedTitleBarModule from './create-need-title-bar.js';
 import posttypeSelectModule from './posttype-select.js';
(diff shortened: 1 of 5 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/components/create-post.js
```diff
@@ -99,19 +133,25 @@ function genComponentConf() {
...
         </div>
-    `
+    `;
 
 
     class Controller {
         constructor(/* arguments <- serviceDependencies */) {
             attach(this, serviceDependencies, arguments);
 
+            //TODO debug; deleteme
```
heikofriedrich: should this be deleted before merging?

heikofriedrich:
i would remove this "add more detail"/"close detail" buttons and just show the location and tags buttons directly there below

heikofriedrich:
the tags input field feels a bit strange. this little boxes where the tags appear thereis always an emtpy one visible. also you can enter tags with or without # character. If you enter text without, its just omitted which also feels a bit strange

Issue #1231, opened 2017-10-02
Solving error message in chat after closing conversation


maxstolze:
When user_1 closes the message connection with user_2, the following error appears as message for user_2:

WonMessageProcessingException: java.security.SignatureException: Could not verify message https://node.matchat.org/won/resource/event/nun0xzk2wglbwgmbuoni: No public key found for https://node.matchat.org/won/resource/event/5386950924531300000#data-sig

maxstolze:
![close_connection_bug](https://user-images.githubusercontent.com/32457902/31127053-010f7ce8-a84e-11e7-80b1-279aa5f11795.png)

maxstolze:
No problem with the closing operation with the following steps:

1.  Close the chat
2. Wait +/- 5 seconds
3. Close connection

-> Own Connection is closed, message gone and chat of the other user also dissapears, connection is closed, message gone.
But: No information for the "other" user, why his conversation is gone.

Log of the "other user" when his chat dissapears

![wait 5 secs](https://user-images.githubusercontent.com/32457902/31337024-caec7ac0-acf9-11e7-8de6-03ca4f5f8337.png)

maxstolze:
Closing connection problem occurs again:
![new_error_1](https://user-images.githubusercontent.com/32457902/31928903-26e6e404-b89a-11e7-8ad2-ad0a9a01615a.png)

maxstolze:
It seems to be a state problem: The two owners aren't updated by the node and try to close the other/remote connection in their "local" states. This connection isn't event set

![state1_1](https://user-images.githubusercontent.com/32457902/31947258-2b345390-b8d4-11e7-85e4-566698c81f32.png)
![state1_2](https://user-images.githubusercontent.com/32457902/31947261-2cb37fac-b8d4-11e7-8fff-86511d6b6f33.png)

quasarchimaere:
as far as i am concerned the "MESSAGE WASN'T PROCESSED...." warning occurs if we receive a message that we do not specifically handle in the ownerwebapp yet, so i assume that implementing this message type and dispatching the event to the state accordingly will resolve this issue

PR #1266, opened 2017-10-17
Fixes the closing connection bug - #1231

maxstolze:
Now the remoteConnectionUri is stored on the sender side to close the connection on both sides.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/reducers/need-reducer.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/reducers/need-reducer.js
@@ -140,6 +140,15 @@ export default function(allNeedsInState = initialState, action = {}) {
         case actionTypes.messages.open.failure:
             return changeConnectionState(allNeedsInState,  action.payload.events['msg:FromSystem'].hasReceiver, won.WON.RequestReceived);
 
+        case actionTypes.messages.open.successRemote:
+        case actionTypes.messages.connect.successRemote:
+            // use the remote success message to obtain the remote connection uri (which we may not have known)
+            var wonMessage = action.payload;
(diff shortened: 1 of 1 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/reducers/need-reducer.js
```diff
@@ -140,6 +140,15 @@ export default function(allNeedsInState = initialState, action = {}) {
...
             return changeConnectionState(allNeedsInState,  action.payload.events['msg:FromSystem'].hasReceiver, won.WON.RequestReceived);
 
+        case actionTypes.messages.open.successRemote:
+        case actionTypes.messages.connect.successRemote:
+            // use the remote success message to obtain the remote connection uri (which we may not have known)
+            var wonMessage = action.payload;
+            var connectionUri =  wonMessage.getReceiver();
+            var needUri =  wonMessage.getReceiverNeed();
+            var remoteConnectionUri = wonMessage.getSender();
+            return allNeedsInState.setIn([needUri, 'connections', connectionUri, 'remoteConnectionUri'], remoteConnectionUri);
```
quasarchimaere: i have some issues with this solution:
-if we add/set the remoteConnectionUri like this we probably produce entries in the connections that only have remoteConnectionUri as the data -> which in return breaks the views that rely on having the correct connection data -> please make a failsafe check to see if the connection is already in the state and then set the remoteConnectionUri accordingly
- Please also update the addConnectionFull method so that it represents the remoteConnectionUri as soon as the connection is added to the state (if the remoteConnectionUri is not present at that given time initialize it with undefined and set it later e.g. here)

quasarchimaere:
did not merge this PR since there are possible issues in the code (see comment)

PR #1270, opened 2017-10-18
Feat auto close

fkleedorfer:


heikofriedrich:
probelms with email feature, will not merge yet

fkleedorfer:
Set the interval to check every 10s, warn after 25, deactivate after 45 on integration-test, which makes for a fun dynamic feel but notifications that have to be clicked away are a bit too much in that setting.

Somehow on that system, the email notifications do not seem to work for me. Could someone else try this? Just sign up, create a need, then sign out and wait for an email. @quasarchimaere @heikofriedrich

PR #1275, opened 2017-10-19
Enable remember-me in owner webapp

fkleedorfer:
* changes signin request from post to get so we can easily pass parameters that can be
picked up by the RememberMeServices

* configures RememberMeServices in owner-security.xml

* changes login method in won-utils.js

* changes the session timeout to 1 day

How to test:
1. log in with an account
2. note there is a 'remember me' checkbox in the login popover
3. do not check that box
4. when logged in, close the window (and all other windows, so the session is closed)
5. open a new browser window
6. open the owner webapp
7. notice that you are not logged in.
8. log in again, now checking 'remember-me'
9. when logged in, close all browser windows
10. open a new browser window and open the owner webapp
11. check that you are now logged in

Note: the remember-me function survives an owner-webapp server restart

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/WEB-INF/web.xml
+++ b/webofneeds/won-owner-webapp/src/main/webapp/WEB-INF/web.xml
@@ -77,9 +77,9 @@
 	    <welcome-file>index.html</welcome-file>
     </welcome-file-list>
 
-    <!-- for testing closing websocket connection purposes -->
-    <!--session-config>
-        <session-timeout>1</session-timeout>
-    </session-config-->
+    <!-- session timeout -->
+    <session-config>
+        <session-timeout>86400</session-timeout>
+    </session-config>
 
 </web-app>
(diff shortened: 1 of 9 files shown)


heikofriedrich:
i think we should be careful with putting the password in the http request parameters. See here for example: https://stackoverflow.com/questions/323200/is-an-https-query-string-secure

webofneeds/won-owner-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -77,9 +77,9 @@
 	    <welcome-file>index.html</welcome-file>
     </welcome-file-list>
 
-    <!-- for testing closing websocket connection purposes -->
-    <!--session-config>
-        <session-timeout>1</session-timeout>
-    </session-config-->
+    <!-- session timeout -->
+    <session-config>
+        <session-timeout>86400</session-timeout>
```
heikofriedrich: 1 day is quite long, i would set this shorter. what about 2 hours?

heikofriedrich:
2 other issues: 
* when you reopen the browser and owner url you come to the page where you can post a new needs about do not see your posted needs, matches etc. => i think this should be another start page for an aldready logged in user
* when you log in and click "remember me" then you close the browser and reopen it you are still logged in => ok. If you post a need now and then close the browser again (without loggin out) and then open it again you should still be logged in but actually you are logged out then

fkleedorfer:
@heikofriedrich your first point is an edge-case we'll have to deal with in the future. The second one has to do (I think) with the browser evicting the cookie because it is marked as 'secure' but the integration-test server only has a self-signed certificate. We'll see if that persists on live.

Issue #1271, opened 2017-10-19
Avoid hints for inactive needs


maxstolze:
Check need state and filter (Active/Inactive) before sending hints.

PR #1282, opened 2017-10-20
Avoid hint for inactive needs

fkleedorfer:
Check the state of the need before sending hint.
issue #1271
(cherry picked from commit e394948)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/messages-actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/messages-actions.js
@@ -290,9 +290,10 @@ export function hintMessageReceived(event) {
         if (!getState().getIn(['needs', event.getReceiverNeed()])) {
             console.log("ignoring hint for a need that is not ours:", event.getReceiverNeed());
         } else {
-            //check the state of the need and ignore hint if it is inactive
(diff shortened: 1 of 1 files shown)


maxstolze:
Is not working with the LastSeenNeedsMatcherApp

maxstolze:
Maybe it is possible to set the "MatchingBehavior" entry to "DoNotMatch" when we close a need to make sure there are no more hints coming from the node for the current need. And when we reopen a need, we set it back to "Mutual".

quasarchimaere:
@maxstolze, this approach is not going to be possible, the problem with this approach is that the changing the behaviour in the need-document would result in a loss of information of the previous matchingbehaviour state, we would have to add an additional previousMatchingBehaviour to the document to ensure that we can restore the previous state, or we would need to crawl through all the messages/events to reconstruct what the state was which in this case would mean that we would have to have unnecessary calculations for those edge cases.

However maybe we need to adapt the matcher in a way that it will not consider closed/inactive needs for matching, this is the most feasible approach as we already parse some need-information for matching anyway (e.g. matchingBehaviour)

maxstolze:
@quasarchimaere +1

heikofriedrich:
what we gonna do with this pull request now? who is working on that?

Issue #1222, opened 2017-09-28
Text-clashes in title-bar


m0ru:
![title clashing with 'new post' clashing with username](https://user-images.githubusercontent.com/1730492/30965359-784e48e0-a455-11e7-95a8-b40c578639c0.PNG)

The responsiveness-break-point for these should be a bit earlier. Also if the username is very long, we might want to hide parts of it, e.g. "johnath...@match....org" or "johnathan.doe" (i.e. hiding the domain). Or we could just hide the user-name based on it's length using js or set a max-width for that element and set text-overflow, e.g.:

```css
max-width: 7rem;
overflow: hidden;
text-overflow: ellipsis;
```

maxstolze:
Now parts of the username are hidden in case it is longer than 16 chars. So there is no clash anymore.
We just need to decide, what we want to hide. At the moment the "middle" part of the name is replaced by "...".

![long_name](https://user-images.githubusercontent.com/32457902/32317599-8f368aac-bfb4-11e7-9d5f-80b98a16e3bd.png)

PR #1295, opened 2017-11-02
Avoid text-clashes in title-bar

maxstolze:
Issue #1222
If the username is longer than 16 characters the middle part is hidden and replaced with "...". Now there are no clashes with the name anymore.


quasarchimaere:
personally i would prefer to use a css based text-overflow: ellipsis solution instead of this javascript approach, however since ellipsis only cuts at the end of the string i think this solution should also be fine... however i would move the method to a util file (preferably utils.js) so we can use it without having to duplicate the code -> preferably a method with two params, the given string and the charactercount when we need to ellipsize

Issue #1286, opened 2017-10-24
Add link to RDF for experts


m0ru:
Technicians evaluating the prototype will want to see the RDF. `won.debugmode = true` only works for us as it's a hidden feature.

m0ru:
This should be done in a way that doesn't let non-technicians end up on the RDF-page.
It should be clear to people what the link / mode does and if they want to use it or not.

PR #1302, opened 2017-11-03
Provide links to the RDF page

maxstolze:
Issue #1286
Now the users can see the RDF pages of their needs and if they want also for their messages.


quasarchimaere:
added minor stylechanges (button with label and image is now aligned vertically

question: would it not be better to add the target blank attribute to the post-info link as well?
question: should we remove the already present debug links from the views as we have the rdf option in the gui anyway?

fkleedorfer:
@quasarchimaere +1

PR #1306, opened 2017-11-09
Update draft-stating-information-requirements.md

fkleedorfer:
Updates the design issue based on conceptual development


quasarchimaere:
i merged this since it is just a draft anyway, so we can change the description/definition on a later date if we choose to do so -> hence why i will leave the branch and not delete it

PR #1307, opened 2017-11-09
Crawl modified needs and connections

### LINKED ISSUE #1292: Matcher does not get informed about need and connection updates
the crawler currently crawls the won node that uses the creation date of needs (and connections => e.g. creation date of first event in connection) to resume crawling at some point. however needs and connections that were already crawled and modified later currently wont be crawled again this way. One solution would be to query all modified needs/connections since a certain date and also crawl them

heikofriedrich:
crawler of matcher is now able to get need and connection updates in every of its crawl cycles (default every 300 min and at startup). crawler uses the etag/if-none-match http headers stop stop crawling at resources with versions it already knows. 

Proposal for testing:

* After needs/connections have been changed trigger crawling in matcher (e.g. by just restarting the matcher)
* the changes (e.g. close/open connections, activated/deactivated needs) should be in the matchers triple store after crawling was successfull

fixes #1292


quasarchimaere:
blazegraph seems to show that the updates after a matcher restart are not saved in the triple store -> see uri https://satvm05.researchstudio.at/won/page/need/6257346360320575000

heikofriedrich:
@quasarchimaere hmmm, for me it is working. i just tried on integration-test, maybe we can try together tomorrow?

Issue #1303, opened 2017-11-06
Closing a need does not close its connections


maxstolze:
When a need with an open connection is closed by a user, the connection at the other user is still open and active. We need to close all connections of a need, when it is closed.

PR #1309, opened 2017-11-13
Fix closing need and connections #1303

maxstolze:
Solves issue #1303

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
@@ -357,10 +357,8 @@ export function needsClose(needUri) {
 
             //Close all the open connections of the need
             const affectedConnections = getState().getIn(['needs', needUri, 'connections']).map(conn => conn && conn.get("uri"));
-            for (con in affectedConnections.toJS()) {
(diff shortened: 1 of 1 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
```diff
@@ -354,6 +354,15 @@ export function needsClose(needUri) {
...
             }));
+
+            //Close all the open connections of the need
+            const affectedConnections = getState().getIn(['needs', needUri, 'connections']).map(conn => conn && conn.get("uri"));
+            for (con in affectedConnections.toJSON()) {
+                if( getState().getIn(['needs', needUri, 'connections', con, 'state']) === won.WON.Connected) {
+                    dispatch(actionCreators.connections__close(con),
+                    )
+                }
+            }
```
quasarchimaere: wouldnt it be better to dispatch the connections__close directly inside the .map function instead of getting all the affected connections making a mutable entity out of it and dispatching the action in a loop?

maxstolze: If this is possible, yes we should do it like that. I did not think about it.

Issue #1300, opened 2017-11-02
fix and activate unit tests in build again


heikofriedrich:
* couple of tests need to be fixed
* tests can be activated on master environment (and/or on integration-test)

PR #1313, opened 2017-11-20
Fix buildwithtests

quasarchimaere:
it should be possible to build the whole project with "mvn install" instead of having to start with the skip-tests profile


heikofriedrich:
realted to #1300

PR #1318, opened 2017-11-24
Goal concept

heikofriedrich:


fkleedorfer:
I am not sure about the use of blank nodes for goals in the taxi example. We may want to reference the goal later on, e.g. when we create a proposal from it, or to reference it in connection with the validation result. My feeling is that it should be a named node, something like `:goal1`, or maybe `:taxiRideGoal`

fkleedorfer:
`event:event1 agr:propose :pickup-solution .` is wrong.

According to the spec (in the agreements paper) the proposal is made by proposing other messages, not graphs inside the same message:

`msg:event1 agr:propose msg:event2.`

This means that you have to send two messages for making a proposal, one containing the data you want in the proposal, and one containing the proposing triples. We've defined that the message containing the proposing triples itself cannot be part of the proposal (i.e. its content graphs are not added to the data that becomes the content of the proposal). We could look if we want to relax that, but there was some reason to disallow the proposing message's content into the proposal.

heikofriedrich:
updated document

heikofriedrich:
added information requests example, ready for review

PR #1342, opened 2017-12-18
Fix gui wording

### LINKED ISSUE #1340: need-textfield.js - more expressive placeholder
in the create-post-view atm the placeholder only reads "what". it might be more expressive and useful to have something like

> **what? - title**
> longer description `#tag`

that shows off the features of the textfield.

fkleedorfer:
Minor changes to the wording of the owner webapp.

* change title of create view to `"Hello, everyone, ..."`
* change placeholder text in create view so it explains better what the create textview does
* change layout to make the textview bigger so it contains the longer placeholder nicely

Fixes #1340

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-need-title-bar.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-need-title-bar.js
@@ -9,7 +9,7 @@ function genComponentConf() {
                 <a class="cntb__inner__left clickable" ng-click="self.back()">
                     <img  src="generated/icon-sprite.svg#ico27_close" class="cntb__icon">
                 </a>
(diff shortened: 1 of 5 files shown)


m0ru:
i think we might have a wot/tldr issue here:

wording variants

^ original (1st), pr as is (2nd), something else (3rd), something shorter (4th)

all of these probably also require another change: ["Placeholders in Form Fields Are Harmful"](https://www.nngroup.com/articles/form-design-placeholders/) (Katie Sherwin on NNGroup, 2014/05). We probably should move the label to outside of field permanently or animate them to move as soon as people select the textfield. for the original version this problem is probably less pronounced, as it's only one word ("what"). see below for alternative designs:

alternative labeling styles

Issue #1345, opened 2017-12-19
Improve geocoding


fkleedorfer:
According to https://operations.osmfoundation.org/policies/nominatim/, we have to cache geocoding results in the client, multiple requests for the same coordinates are considered faulty and blocked. 

Blocked geocoding queries are responsible for unresponsiveness in the owner app. 

Two consequences for us: 
1. As we cannot assum the service to be available at all times, the app must be more forgiving in such cases. 
2. We really should cache query results in the client.

maxstolze:
So far the geocoding works fine with the minimal solution (use browser coordinates in case of a bad response).

PR #1348, opened 2017-12-21
Improve geocoding and fix openstreetsmaps problem

maxstolze:
issue #1345
If the openstreetmaps server is not reachable, or blocking a request,
the current location from the browser is taken to create a "what's
around" need and the boundingbox is computed with the "geopoint"
library, which is now included in the project.


m0ru:
there seems to be something white-space-related in the diff of jspm_config.js and package.json (probably unix vs windows file-endings)

m0ru:
same goes for the entirety of utils.js

quasarchimaere:
i force pushed the same changes onto this branch, so there should not be a whitespace problem anymore

PR #1366, opened 2018-01-07
Refactor owner webapp

fkleedorfer:
* Optimistically add need immediately after posting

* New chaining mechanism for actions
  Allows for executing actions when a success/failure response for
  a given message is received

* Use chaining mechanism to wait for success of AdHoc need creation,
  then send a connect message

* Refactored use of connect and open (connect requires no existing
  connection

* What's around need creation now only deactivates all existing *active*
  what's around needs, instead of all of them (which is wasteful)

* Removed some excessive log output (still more to be done there)


fkleedorfer:
Deployed on integration-test, ready to merge

PR #1368, opened 2018-01-08
Fix route change blocked #1367

### LINKED ISSUE #1367: Route change blocked

quasarchimaere:
fixes #1367 

problem was that there was a routechange when closing a need, and a route change when a need is created, however creating a whatsaround need closes all preexisting whatsaround needs and due to the asyncronous behaviour of routechanges there was an overlap in the transitions resulting in the described error.

Fix was: remove the routechange in the need close action, and make a routechange for createneed that always goes to the feed when a need is about to be created (thats currently our defaultRoute)


quasarchimaere:
already pushed onto integration-test

PR #1389, opened 2018-01-18
Feat rdfanalyzer

quasarchimaere:
This pull request is just here so we can push the current changes of the rdf_analyzer into master and thus reducing the risk of any merge conflicts (please do not remove this branch after merging)


heikofriedrich:
looks fine to me so far + integration test is still working correctly it seems

PR #1411, opened 2018-02-02
Feat basicskinning

heikofriedrich:
add skins for default, matchat and uki deployments.
already deployed on uki environement, can be tested there


m0ru:
only thing missing is being able to replace the entire about page if necessary (atm that's limited to the impressum)

PR #1443, opened 2018-02-16
Feat agreements in debug bot

fkleedorfer:
Adds new commands to the debugbot for testing agreements

How to test:
1. connect to the debugbot
2. type 'usage' 
(read about the new commands)
3. switch RDF mode to 'SHOW'
4. type 'retract' and hover over the message in which the bot tells you that it 'hereby retracts' a message. You should see that there is a 'mod:retracts' triple in there.
5. try the same with the other new commands.

Note: you can send a custom triple in the chat using `::msg::` to refer to the message's URI. 
For the predicate, the prefixes `won`,`msg`,`agr`, and `mod` are expanded. For example, you can propose a message by sending 
```
::msg:: agr:proposes <https://192.168.124.49:8443/won/resource/event/rlgrjqfrjp1qcsmzip8x>
```


fkleedorfer:
Note: you can send a custom triple in the chat using `::msg::` to refer to the message's URI. 
For the predicate, the prefixes `won`,`msg`,`agr`, and `mod` are expanded. For example, you can propose a message by sending 
```
::msg:: agr:proposes <https://192.168.124.49:8443/won/resource/event/rlgrjqfrjp1qcsmzip8x>
```

heikofriedrich:
when using retract "retract", there is not really the "last" message be retracted but a message that is just to be sent: 

Ok, I am hereby retracting my last message, which read, 'ok, I'll retract my latest message - but 'll need to crawl the connection data first, please be patient.'

fkleedorfer:
@heikofriedrich yes, you're right - that's better

PR #1466, opened 2018-02-28
Matching fix

heikofriedrich:
* do not use dummy query anymore for standard need queries to allow 0 results if query doesnt hit anything
* fix typo in solr field name of seeks.description and seeks.seeks.description

**NOTE:** if this pull request is deployed to live/uki the solr index has to be deleted there so that it can rebiuld


fkleedorfer:
Much better now, I'd say.

PR #1472, opened 2018-03-08
Feat  rdf in chat

m0ru:
Merge of `agreements_in_chat_gui_#1451` and `feat_analyzeaction_agr_prop` with any changes from `feat__rdf-in-chat` to avoid future, more complicated merge conflicts in regard to the need-reducer and the connection-message-component. Please check if everything you're working on still works the same as before, I could only check my stuff as I don't know the state of developement on your branches.

Subsumes `#1471` (PR) by it's nature (either merge that first and then this one, or just this one)


quasarchimaere:
"Subsumes #1471 (PR) by it's nature (either merge that first and then this one, or just this one)" -> please merge #1471 first since it has additional changes to the branch

Issue #1514, opened 2018-03-29
Agreement GUI: I should have option to propose counterpart's messages


fkleedorfer:
I can only propose my messages, but the protocol allows to propose any message. Hence, the counterpart's messages should also be proposeable.

Issue #1512, opened 2018-03-29
Agreement GUI: accepted cancellation should not look like an agreement


fkleedorfer:
This is a tricky one, because a proposal can contain clauses and cancellations. 
These cases are possible when a proposal is accepted:
* it contains only clauses
* it contains clauses and cancellations
* it contains only cancellations
Each of these cases should get its own graphical representation, just saying it's an 'Agreement' is not enough:
![grafik](https://user-images.githubusercontent.com/551624/38081918-83fc80ec-3345-11e8-964d-a424f5b6f347.png)
(I also copied the debugbot's eliza message because it's so fitting here ;-) )

Issue #1511, opened 2018-03-29
Agreement GUI: agreement controls should always be visible 


fkleedorfer:
It is easier to understand how agreements work if the buttons are visible.

This:
![grafik](https://user-images.githubusercontent.com/551624/38081538-941eed4e-3344-11e8-988f-0cda903effea.png)

...is more easily understood than this:

![grafik](https://user-images.githubusercontent.com/551624/38081624-c86269d2-3344-11e8-9f66-ec64f1547dae.png)

Issue #1510, opened 2018-03-29
Agreement GUI: agreements should not be called 'propose to cancel'


fkleedorfer:
An agreement is still an agreement when someone proposes to cancel it. It only ceases to be an agreement when the cancellation is accepted. We currently call these 'cancelled agreements'.

In this example, we have 2 agreements, both proposed to be cancelled.
![grafik](https://user-images.githubusercontent.com/551624/38081344-1a75d110-3344-11e8-83fd-43725f4113e6.png)

PR #1523, opened 2018-04-06
Refac propose to cancel

maxstolze:
Solve proposeToCancel Problem + debug DebugBot
 -> Solve #1508 #1510 #1511 #1512 #1514

No retract and reject now in this pull request


fkleedorfer:
When clicking 'show Agreement Data' after proposing/accepting something, the GUI panel often takes much longer to be updated and show the agreement data than the call to getAgreementProtocolUris takes. I did not time it exactly, but the service call takes 1.9s, the panel only updates after 60s or so. Sometimes, the update is instant, though.

PR #1527, opened 2018-04-09
refactors the random number generator so that the first symbol of a g…

quasarchimaere:
…enerated string is never a number (without the use of a loop)


quasarchimaere:
i didnt implement this to be nitpicking, i merely saw that this code snippet would (while it still works) call the generation roughly 25%  more often than without the loop which is not really optimal since these id's are generated fairly often

fkleedorfer:
+1

PR #1528, opened 2018-04-09
refactors newMessage and newConnection parameters in state to "unread…

quasarchimaere:
…" and add "unread" to the need in the state as well.

this is so we can make sure to add the won-unread class to needs as well....

additionally the need and connections will be set to read if there are no unread subelements left (e.g a need is unread=false if there are no connections with unread=true within this need, and the same goes for connections and their respective messages... for connections that have the state "connected" we will not set them to read immediately after clicking on the element anymore


quasarchimaere:
i will close this pull because the changes are included in a different pr now

PR #1538, opened 2018-04-13
Feat updatepostview  and  link

### LINKED ISSUE #913: Inc Request message in same style as chat messages
We should be consistent and as simple as possible with our style / mapping from style to semantics. They're displayed as chat messages later and basically have the same conversational status as chat messages (as all of them are private communication between the same two persons, as opposed to need descriptions). Using two different styles for essentially the same thing can cause confusion.

quasarchimaere:
show postings without connections in chats view, too~~ (with an indicator panel saying 'This posting has no connections. Consider posting this link (link) in social media, or wait for matchers to connect you with others.

make popdown on needs in chat view that shows all details (own and remote). Popdown should activate on title click or image/icon click. Also add an option to reopen closed needs.

fixes #913


webofneeds/won-owner-webapp/src/main/webapp/style/_post-messages.scss
```diff
@@ -104,9 +104,18 @@
         padding: 0.5rem 0;
         min-height: 3.25rem;
       }
+
+      &__labelledhr {
```
peacememories: Empty selector, remove.

Issue #1532, opened 2018-04-10
Agreements GUI: don't show agreement-related controls until agreement protocol state is known


fkleedorfer:
Steps to reproduce:
* create a conversation with an agreement
* make sure the proposal one of the latest messages so it is visible in the chat window when scrolled down completely
* reload the browser window
-> observe that the proposal message shows an 'accept' button that vanishes as soon as the APS is loaded

It would be better to hide all such controls until we know we need them, instead of to show them by default and hide them when we know we don't need them.

Issue #1530, opened 2018-04-10
Add "isRelevant" to chatMessage


maxstolze:
Add a new flag "isRelevant" to message in state:

isRelevant: true -> message is shown in chat
isRelevant: false -> message is crossed out in chat, no interaction possible anymore

PR #1542, opened 2018-04-13
Add isRelevant

maxstolze:
Solve #1530 and #1532


quasarchimaere:
when a proposal is accepted by the other need, it seems like the isRelevant status is not updated -> meaning proposals that have been made are not marked as isRelevant=false -> PR can't be merged yet

maxstolze:
fixes #1532

Issue #1322, opened 2017-11-28
Make match-detail popup full-screen on smaller devices


m0ru:
![matches view on small screen](https://user-images.githubusercontent.com/1730492/33332277-2a19f68a-d464-11e7-8f6b-99509d6353fa.PNG)
This should be **fullscreen** or at least **vertically centered** according to the actual height of the screen. For the latter: use something like `calc(50vh-50%)`(?) or define a 3x3 grid with outer colums/rows that can become 0.

PR #1544, opened 2018-04-13
Feat update post visitor view

sigpie:
Changes visitor view - to test, check post visitor view and all places where won-send-requests is used.


sigpie:
Closed #1322

Issue #1545, opened 2018-04-13
Responsiveness Breakpoint - declining match does not return to connection overview


sigpie:
Closing a hint with the "Bad match - remove!" button should remove the connection and display the list of remaining connections. Instead, the connection is removed and nothing is displayed.

PR #1557, opened 2018-04-13
Fix #1556

### LINKED ISSUE #1556: Getting stuck in 'right hand side' of chats on small devices
Steps to reproduce:
1. Use a phone in portrait orientation
2. Do something to get matches (e.g. What's Around)
3. navigate to the chats view 
4. click on one of the matches or click the matches indicator button. Result: you now see the detail view of the matched need
5. click 'bad match - remove'

Result: you only see a grey area, the text 'No Element selected' is not shown either. changing device orientation or clicking 'chats' in the navigation bar at the top gets you back to the chats view.

peacememories:
Now no longer gets stuck on gray screen.

fix #1556 #1545

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/feedback-grid.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/feedback-grid.js
@@ -70,6 +70,7 @@ function genComponentConf() {
                     console.log("RATE BAD");
                     this.connections__close(this.connectionUri);
                     this.connections__rate(this.connectionUri, won.WON.binaryRatingBad);
+                    this.router__stateGoCurrent({connectionUri: null})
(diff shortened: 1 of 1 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/components/feedback-grid.js
```diff
@@ -70,6 +70,7 @@ function genComponentConf() {
                     console.log("RATE BAD");
                     this.connections__close(this.connectionUri);
                     this.connections__rate(this.connectionUri, won.WON.binaryRatingBad);
+                    this.router__stateGoCurrent({connectionUri: null})
```
peacememories: Maybe this route change should be dispatched in the connections__close action if the current route matches the connection route?

quasarchimaere:
also fix #1545

PR #1558, opened 2018-04-16
Feat  remove post view #1521

quasarchimaere:
fix #1521, removed post-owner-view and all links to the post-owner page as well as refactored the old overview-incoming-request urls to connections

missing: if you have a visitorlink but own the post an automatic redirect to the connections page is not present yet

PR #1559, opened 2018-04-16
Fix  #1547

### LINKED ISSUE #1547: Post-info link in footer should display correct address
The post link shows the resource url, not the userfacing url.

quasarchimaere:
fixes the post-info link for suggestions and need-info (url is currently built with window.location and a fixed webapp part... i am sure this could be enhanced)

UX updates should be in a different/separate branch and i will create a ticket for that


quasarchimaere:
blocked by #1558

PR #1616, opened 2018-04-23
filter all closed whatsaround needs from connections-overview -> issu…

### LINKED ISSUE #1588: Old and inactive whatsaround needs stay in overview
Every time a new "See What's Around?" need is created, the previous one is set to inactive. If this happens several times, old "See What's Around?" needs accumulate in the overview. 

closed_whatsaround_needs

It would be better to show no or only the most recent inactive "See What's Around?" need in the overview instead, especially since there is no reason to reopen old "See What's Around?" needs.

quasarchimaere:
fixes #1588


fkleedorfer:
Does not change anything - this screenshot was taken after logout + login

PR #1635, opened 2018-04-24
Fix #1617

### LINKED ISSUE #1617: Show unread counter on own needs in single-column view
on a small (mobile) device, the match/message icons on the needs are removed. This hides important information. We should at least show an unread counter.

quasarchimaere:
fixes #1617 shows unread counter for all unread connections in mobile view


fkleedorfer:
Does not work for incoming connections or matches:
grafik
... but there is one match and one request:
grafik

quasarchimaere:
i agree that this isnt working and apologize, after reviewing the changes i see that i pushed the wrong branch

quasarchimaere:
wrong branch

PR #1641, opened 2018-04-24
Add is relevant #1530

### LINKED ISSUE #1530: Add "isRelevant" to chatMessage

### LINKED ISSUE #1532: Agreements GUI: don't show agreement-related controls until agreement protocol state is known

maxstolze:
Now all accepted proposals + propose to cancel are set to isRelevant = false after accepting them

Fixes #1530
Fixes #1532
Fixes #1542


maxstolze:
Blocked by #1643

maxstolze:
Still blocked by #1643

PR #1645, opened 2018-04-24
Fix  #1598

### LINKED ISSUE #1598: Don't send hints to closed needs
Our matcher sends hints to closed needs. That should not be the case.

peacememories:


fkleedorfer:
Does not fix the problem. Steps to reproduce:
1. Create need A with user X
2. Create need B with user Y, making sure it will match A
3. observe that they are matched and you can connect.
4. with user X, deactivate need A.
5. with user Y, create need C that would match A
6. observe that Y is matched to A.
Verify that A was really matched:
7. Reactivate A
8. connect to matched need on need C
9. observe that A gets an incoming request.

PR #1649, opened 2018-04-25
forgot " @ mobile indicator class attribute

### LINKED ISSUE #1617: Show unread counter on own needs in single-column view

quasarchimaere:
fixes #1617


fkleedorfer:
Does not fix the issue. 
Reproduce:
1. activate single-column layout
2. In debug mode, create a need
3. wait for hint/connect
4. this is what the overview looks like:
grafik

but there are new events:
grafik

PR #1650, opened 2018-04-25
Work around for copy/paste problem

fkleedorfer:
issue #1269
Try to find a workaround for the copy/paste problem in the chat text field


fkleedorfer:
@maxstolze 
Is this still something we need? If not, please close the PR and delete the branch.

quasarchimaere:
this is not relevant anymore -> we switched from content editable to textarea

PR #1677, opened 2018-04-27
Feat limit and restyle unreadcounters #1554

### LINKED ISSUE #1554: Unread-number on indicators can break layout.
If the unread-count on an indicator in the chat view is more than one digit long it nudges the whole line to the left.

image

Possible solutions:

1. Make the minimum width `n` characters (`n` would probably be two-ish). This would break the next order of magnitude up.
2. Do `1.` but have a maximum number of displayed notifications and just show something like "99+" after that.
3. Remove the number beside the indicator and add it as a badge on the indicator (eg. Macos Dock badges ). Legibility is a concern here!

### LINKED ISSUE #1685: unread counter '99+' when it shouldn't
In PR #1677

This seems to happen under at least two circumstances:
A: cliking on message indicator after incoming connect.
Reproduce: 
1. create need in debugmode
2. wait for connect/hint from bot
3. click on link indicator. 
Observe that the number next to it will show 99+ for a moment.

Testing this more shows that the number is 99+ most of the time - I don't think more reproduction instructions are needed - it seems that when nothing is unread, it shows '99+'

quasarchimaere:
default limit of any connectionCount in the connectionindicators is 100 if the limit is reached we will just use "99+" instead

fixes #1554 
fixes #1685

also changes some ng-show to ng-if to remove a jumping view when animating some elements


fkleedorfer:
Blocked by #1685

quasarchimaere:
retest possible, bugfix for #1685 was included in this branch

PR #1695, opened 2018-05-03
Feat style headers

quasarchimaere:
this pull request changes the visible information of the connection headers (subtitle is now the connection state instead of what the remoteneed is)
additionally i changed the layout to a grid and will ellipsize anything that overflows from the title (includes white space no wrap)

therefore all the headers will be the same height


fkleedorfer:
Blocked by conflicts

PR #1710, opened 2018-05-08
fixes issue with firefox layout - changed flex to grid to ensure usin…

### LINKED ISSUE #1704: Long titles mess up layout
In firefox, long titles aren't handled properly:
wide screen seems kind of ok, but still weird because the title is not abbreviated, it just overflows into an invisible ara:
grafik
less space: not so good
grafik

quasarchimaere:
…g the whole available space even for short titles, and set min-width to zero so that even firefox will 
collapse the child elements

fixes #1704 

i changed the display property from flex to grid to ensure that the child elements use the whole available space, because, min-width: 0 is necessary so firefox collapses grid (or even flex) and not display the whole child element -> this is rather unituitive but well explained in this link https://bugzilla.mozilla.org/show_bug.cgi?id=1108514


quasarchimaere:
included in #1711

PR #1732, opened 2018-05-09
Fix jumping columns #1729

### LINKED ISSUE #1729: Columns "jumping" when post subtitle is too long
When the post subtitle is too long (e.g. "Post (no matching context specified)", it get's wrapped next to the connection indicators. As the indicators disappear when the post is opened, the text gets unwrapped and the columns "jump" to adjust to the new width.

post_closed
post_open

quasarchimaere:
fixes the "jumping" columns by also abbreviating the subtitle in favor of breaking into multiple lines

fixes #1729

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/connections-overview.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/connections-overview.js
@@ -146,7 +146,7 @@ function genComponentConf() {
                         </won-extended-connection-indicators>
                         <div class="co__item__need__detail__actions">
                             <button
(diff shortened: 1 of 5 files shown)


peacememories:
The original issue does not seem to be completely resolved:

image
image

The overview column changes size drastically when folding and unfolding needs.

quasarchimaere:
should be resolved due to a min-width of the list in the non mobile view

peacememories:
The jumping is fixed, but while animating the "Show Post Details" button button is cut off on the right:
image
Also the horizontal padding jumps slightly when hitting the responsiveness breakpoint.

peacememories:
Moving the remaining problem to a separate issue and merging this.

PR #1749, opened 2018-05-11
Bumped n3 version to 1.0.0-beta.1

peacememories:
Contains fixes for https://github.com/rdfjs/N3.js/issues/135 and https://github.com/rdfjs/N3.js/issues/136

This allows us to minify function names again, and makes it possible to use synchronous parsers in the future.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/package.json
+++ b/webofneeds/won-owner-webapp/src/main/webapp/package.json
@@ -14,7 +14,7 @@
     "leaflet": "0.7.7",
     "loadcss": "0.0.2",
     "medium.js": "1.0.1",
-    "n3": "1.0.0-alpha",
+    "n3": "^1.0.0-beta.1",
     "ng-redux": "3.4.1",
     "rangy": "1.3.0",
     "reduce-reducers": "0.1.1",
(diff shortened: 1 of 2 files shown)


quasarchimaere:
i would not merge this pr for now because the version number seems to be a bit fidgety and could cause problems if devs have different versions of the lib, we should rather fix it to the latest version that we know exists and continuously update it

peacememories:
The version number we have right now is `1.0.0-alpha` so I don’t think the new version is more fidgety

m0ru:
^ what he says

The reason for bumping to 1.0.0 in the first place was, that it uses the new rdf-js standard for serialising data, which substantially improves interop between the different rdf-related libraries we use.

quasarchimaere:
oh i didnt mean that i am not ok with the version in general, i just meant that we should just remove the ^ prefix from the version number, since we had issues with different versions on different build systems in the past and this is a pita to find and fix if it occurs

m0ru:
aaaah, lol, sure ^^

quasarchimaere:
wait why was this pr merged with the fluent versioning of the dependency now?

peacememories:
This may not have been communicated clearly:

With the newer npm versions we have access to package locking now. The
`package-lock.json` specifies the fixed versions of every package (even
transitive dependencies) we use and makes sure that on an `npm install`
those exact versions get installed. So there is no need anymore to specify
fixed dependencies in our `package.json`, and since version ranges are the
default result of `nom install <package>` and give us the ability to do
guided updating I recommend we stick with that.
There might already be an issue open to convert all dependencies to range
dependencies.

PR #1783, opened 2018-05-18
fixes #1765 quick fix for \n in description (also a quick fix for \n …

### LINKED ISSUE #1765: post-info: description text needs to parse `\n`s into `<br/>`s
image

quasarchimaere:
…in messages) -> we do not replace the \n with <br> but just preformat the output element

fixes #1765 

additionally allows us to display newlines within sent textmessages
refactors the prewrap style of the code element in order to remove the indentless lines in the template (prone for restyling errors) by moving the prewrap attribute into the specific code tags and not the parent element, thus creating an even neater look than before as we lose the unnecessary newlines at the start of the rdf content in a connection-message


quasarchimaere:
i chose this "quick fix" in favor of a solution where we preformat the text into paragraphs because it would be less change to implement (especially in the description of a post, as these are already encapsulated into paragraphs (for whatever reason))

PR #1792, opened 2018-05-22
Refac  add tags picker component

### LINKED ISSUE #1580: Tag-textfield not in sync with tags
image

The text-field should be in sync with the added tags.

As this is a rather hard problem to handle (in particular, deleting a tag from the synched text-field should also delete / untaggify it from the main text-area or exclude it from the tag-list), it might be better to side-step the problem. e.g. the "add tag" button just adds the `#`-character at the cursor in the main text-field (or sets the cursor there if it isn't already). or we could defer to just handling tags in the separate text-field. or something else entirely.

### LINKED ISSUE #1745: Edge-case where tags aren't removable
1. add tags using the tag-input-field
2. switch to location picker
3. go back to tag-input-field -- the content is gone but the tag is still there above it and can't be clicked away. the only way to remove them is to remove all tags.

image

might already be fixed but not live yet. @sigpie? :)

sigpie:
Moved tags picker to own component. Tags can now only be set via the tag picker and should not disappear while editing other details. 

Closes #1580, closes #1745 by restricting tag addition/deletion to only one textfield.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-isseeks.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-isseeks.js
@@ -6,14 +6,7 @@ import angular from "angular";
 import "ng-redux";
 import won from "../won-es6.js";
 import { postTitleCharacterLimit } from "config";
-import {
-  attach,
-  deepFreeze,
-  clone,
-  dispatchEvent,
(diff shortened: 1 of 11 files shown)


quasarchimaere:
blocked until master is exposed (currently configuring it) will post the status as a comment, please do not merge it for now

PR #1797, opened 2018-05-22
Made base-uri confable via default.js and live.js

m0ru:
It's not drop-in confable like the skin-folder, but I wanted to avoid introducing another round-trip to the page load and keep the application statically servable.

Note: can't properly test this, as my setup's fubar atm. The work for this PR was an attempt to circumvent this by using an alternative owner-server like integration-testing.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-message.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-message.js
@@ -156,14 +156,14 @@ function genComponentConf() {
             </div>
             <a ng-show="self.shouldShowRdf && self.message.get('outgoingMessage')"
                 target="_blank"
-                href="{{self.ownerBaseUrl}}rest/linked-data/?requester={{self.encodeParam(self.ownNeed.get('uri'))}}&uri={{self.encodeParam(self.message.get('uri'))}}&deep=true">
+                href="{{self.ownerBaseUrl}}/rest/linked-data/?requester={{self.encodeParam(self.ownNeed.get('uri'))}}&uri={{self.encodeParam(self.message.get('uri'))}}&deep=true">
                     <svg class="rdflink__small clickable">
                             <use xlink:href="#rdf_logo_2" href="#rdf_logo_2"></use>
                     </svg>
             </a>
             <a ng-show="self.shouldShowRdf && !self.message.get('outgoingMessage')"
                 target="_blank"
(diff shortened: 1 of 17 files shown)


m0ru:
Not mergeable yet. @peacememories and I ran into some probs while testing it on his machine. We'll look at it again tomorrow.

webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-message.js
```diff
@@ -154,14 +156,14 @@ function genComponentConf() {
...
-                href="/owner/rest/linked-data/?requester={{self.encodeParam(self.ownNeed.get('uri'))}}&uri={{self.encodeParam(self.message.get('uri'))}}&deep=true">
+                href="{{self.ownerBaseUrl}}rest/linked-data/?requester={{self.encodeParam(self.ownNeed.get('uri'))}}&uri={{self.encodeParam(self.message.get('uri'))}}&deep=true">
                     <svg class="rdflink__small clickable">
                             <use xlink:href="#rdf_logo_2" href="#rdf_logo_2"></use>
                     </svg>
             </a>
             <a ng-show="self.shouldShowRdf && !self.message.get('outgoingMessage')"
                 target="_blank"
-                href="/owner/rest/linked-data/?requester={{self.encodeParam(self.ownNeed.get('uri'))}}&uri={{self.encodeParam(self.message.get('uri'))}}">
+                href="{{self.ownerBaseUrl}}rest/linked-data/?requester={{self.encodeParam(self.ownNeed.get('uri'))}}&uri={{self.encodeParam(self.message.get('uri'))}}">
```
quasarchimaere: please change the ownerBaseUrl so it does NOT expect a trailing slash at the end, that way the links wont break even if the conf param has a trailing slash at the end

peacememories:
I think the problem we ran into is because our input should look like this:
`import ownerBaseUrl from 'config'`
instead of
`import {ownerBaseUrl} from 'config'`
because we are not deconstructing an object but importing a certain name.

webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-message.js
m0ru: mhm, good point
after trying a few variations, i introduced `urljoin` that handles path-concatenation correctly.

m0ru:
@quasarchimaere  mhm, good point
after trying a few variations, i introduced urljoin that handles path-concatenation correctly.
everything should work now as intended on the client side. still needs some changes to spring's security settings to actually leverage this -- for now it just works as before but with less string-literal duplication.

PR #1814, opened 2018-05-23
Replaced preact with react for compatibility reasons.

peacememories:
DO NOT MERGE

This introduces React into the code base for one component. It also uses styled-components, which could be interesting to us but create a disjointed css experience right now because they can't tap into our scss mixins.


peacememories:
While I still want to bring better component management to the codebase, this approach feels subpar.

PR #1827, opened 2018-05-25
Fixed sourcing of piwik configuration

### LINKED ISSUE #1816: piwik doesnt record actions on matchat.org anymore since 16.05.18

peacememories:
Fixes #1816 

The problem was that `env` in the webpack config does not, in fact, contain the process environment but the webpack environment.

In addition to the fix, i relaxed the config a bit so any value can be passed to WON_DEPLOY_NODE_ENV and the corresponding config script is selected.


heikofriedrich:
ok cool, i checked its working again on live

Issue #1594, opened 2018-04-20
Add option to turn anonymous accounts into normal accounts


sigpie:
As of now, visitors can either create a need or create an account on their first visit. If they create a need, an anonymous account is created and the signup button is hidden. 

As a result,
* there is no obvious way for a new user to create an account after a need has been created
* there is no way for a new user to associate an email address to an anonymous account -> they can't receive notifications via email
* there is no way for a user to associate an anonymously created need their account

Possible solution: keep signup button visible until a normal account is created and give users the option to keep their private id (and all associated needs) on signup.

PR #1828, opened 2018-05-28
Feat account transfer

### LINKED ISSUE #1698: migrate anon to normal account

quasarchimaere:
as soon as there is a privateId in the url the user is able to transfer the private account into a newly registered account

fixes #1698 

ui - is currently not polished, would be subject to a different pr if necessary, ws method names and rest path/method can still be refactored as seen fit in a different pr if necessary


maxstolze:
Resolved #1594

PR #1839, opened 2018-05-30
fixes the disappearing icon when selected (detail picker) #1791

### LINKED ISSUE #1791: Detail Picker icons disappear when selected
The icons used by the details picker should change to checkmarks once the detail is picked, not disappear.  While no value is set for the detail, the icon is partly visible while selected, but not correctly coloured. Once a value is set, the icon disappears on select. 

Expected behaviour: icons are always circled, red when selected, black when not selected. 

mssing_icon

quasarchimaere:
due to reasons unknown to me the scss to css compilation does not replace the $won-primary-color placeholder if it is assigned to a css variable: therefore we need to reference it with the "real" css var value instead of using the syntactic sugar of scss....

fixes #1791

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/style/_create-isseeks.scss
+++ b/webofneeds/won-owner-webapp/src/main/webapp/style/_create-isseeks.scss
@@ -47,7 +47,7 @@ won-create-isseeks {
           //color: $won-line-gray;
           //--local-primary: $won-line-gray; // for the svg-icons
           color: $won-primary-color;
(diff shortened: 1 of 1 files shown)


peacememories:
I fixed the same problem somewhere else in the codebase, while still using the DVD’s variables, I think. I can look it up once I’m at work

quasarchimaere:
i guess we would be able to fix this with #{$won-primary-color}; as well @peacememories

PR #1862, opened 2018-06-04
Feat remove perfomance hogging method

quasarchimaere:
adding log to (almost) every method of linkeddata-service-won.js to ensure we all see how many times and when methods are called

making a first performance update -> remove the unnecessary ensureRdfsMemberArrays from the code base, it was never creating a different object and the recursion went to deep to begin with (the recursion was called for each individual character of a string too -> and thus creating many many unused method recursions )


fkleedorfer:
seems much better on integrationtest. merging and pushing to live

Issue #1625, opened 2018-04-24
Mobile View Input Fields are hidden


quasarchimaere:
Problem: 
- Our input fields are usually on the bottom of the screen (in chat and requests)
- If the app is used on a mobile device the software keyboard opens after tapping an input field and therefore hiding the input field underneath the keyboard until the user types the first character (this is true for android i am not sure if this is the case in ios), idk i think the viewport of the screen is only reduced once the a character is typed, maybe there is a way to add an attribute to the input fields that indicates that the browser should resize the viewport on selection and not after the first character has been typed...

fkleedorfer:
@quasarchimaere @peacememories @m0ru can you think of a way to fix this?

maxstolze:
So far it works in the "chatview", to reduce the grid-template-rows on mobile devices from 10rem to 5rem.

quasarchimaere:
thats great! what about the input in the login dropdown/signup/need creation? did the removal of medium.js and the reduction help there as well?

maxstolze:
Work in progress! I am pretty shure, yes!

PR #1873, opened 2018-06-06
Quickfix mobile view input #1625

maxstolze:
This does quickfix the problem in the chat-view.
The keyboard problem for the other views will be fixed in a seperate pull request.
Part of #1625


maxstolze:
As suggested by @quasarchimaere we should avoid uneccessary code and functions in the quick fix -> grid rows static set to 5rem

sigpie:
Fixes the disappearing chat input field, but does not scroll the chat and does not work for need creation yet.

PR #1886, opened 2018-06-07
Feat enhance connuri crawling

quasarchimaere:
instead of crawling all the connection uris directly and only returning the ones we actually want we just get all the connectionUris from a need, and then storing the closed ones similar to the process we will implement for deactivated needs (in the local storage) so we do not ever crawl a closed connection again on reload/login etc... the connection is marked as closed if the connection is added to the redux state and has a state of WON.Closed, or when the connectionState itself changes

Fixes in this branch:
-removed an unused method
-called the executeCrawlableQuery with the right set of parameters (3 instead of 2)
-omit further crawling for a connUri that has already been deemed as closed (from the localStorage)

This should enhance the login speed for users with alot of needs and connections tremendously in the future (first time and incognito access will not go any faster as we will not know which connections are already closed, but everybody else will hopefully experience a huge performance enhancement on the second reload and or login)

happy happy joy joy

ps: @fkleedorfer please don't be mad, that i did this, i needed a palate cleanser after the gui fixes

One should see the change by looking into the localStorage and finding a set of connectionUris that have already been closed, and also on login/reloads after the first one has been completed you should find the following logentry for connecitons that have been omitted in crawling  "[connUri] closed, ommit further crawl for this uri" (without the [])

calling won.clearClosedConnUris()  in the debug console, resets this localStorage element to nothing again (for testing purposes)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
@@ -91,7 +91,6 @@ const INJ_DEFAULT = "INJECT_DEFAULT_ACTION_CREATOR";
 const actionHierarchy = {
   initialPageLoad: pageLoadAction,
   connections: {
-    fetch: cnct.connectionsFetch,
     open: cnct.connectionsOpen,
     connectAdHoc: cnct.connectionsConnectAdHoc,
     close: cnct.connectionsClose,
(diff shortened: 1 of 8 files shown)


quasarchimaere:
there is still an issue in this one, as it seems, even with these last commits
i will block this pr until further notice

quasarchimaere:
pr can be rechecked, is not yet on integration-test  though

quasarchimaere:
app. there was a commit that was necessary that i seemed to forget to push and therefore it did not work (utils.js) -> last commit fixed a wrong expression

quasarchimaere:
seeomittedconnections
seems to work now...

PR #1895, opened 2018-06-11
change styling (color and mobile positioning) of the ws-sock problem …

### LINKED ISSUE #1872: Make 'Lost Connection' Banner look less like a big problem
Currently, when the websocket crashes, this is what I see:
grafik

This looks like something really went wrong and should never happen. Sadly, this always happens when the browser isn't used for some time, so it is not an emergency state but something quite normal.

Along with other measures that should alleviate this situation (e.g. make reconnecting actually work, see #898 #990 #1789) we could simply give it a different color: let's try the secondary color for the background.

quasarchimaere:
…slide-in #1872 

fixes #1872

changes the background of the conn lost slide-in (to won-secondary-lighter) and the positioning for the mobile view (flex column instead of flex row)

also removes the rather redundant toast that tells you about the exact same problem (see screenshot of the issue)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/topnav.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/topnav.js
@@ -18,7 +18,6 @@ import * as srefUtils from "../sref-utils.js";
 
 function genTopnavConf() {
   let template = `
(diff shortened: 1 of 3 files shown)


fkleedorfer:
After clicking 'reconnect', the animated gif for loading is shown - but the background of that icon is red.

Issue #1866, opened 2018-06-06
Set up MC instance


fkleedorfer:
We can re-use the existing uki instance.
What we have to change:
- [x] wipe data
- [x] skin
- [x] change server name: we'll host an instance under matchat.org (meicogsci.matchat.org)
- [x] context and config

Issue #1669, opened 2018-04-26
Update About Page


quasarchimaere:
the current state of the about page is pretty useless to say the least...we use terms that are not up-to-date and icons that don't exist anymore... we should either update this page or remove in all together in favour of putting focus on implementing an onboarding-bot, i prefer the latter

PR #1915, opened 2018-06-13
Feat disclaimer slidein #1888

### LINKED ISSUE #1888: Add privacy policy etc. links to main page
We need to link to a few documents from the main page. 
Ideally, this is entirely configurable in the skin, such that we can have different numbers of links and entirely different documents per skin

child of #1866

use the imprint.html (within the uki skin (meicogsci) to add all necessary content in terms of agreement etc.)

then implement a slide-in (similar to the slide-in for reconnect (is implemented in topnav.js) that pops up if a imprint.html is present within the skin and a certain flag "privacyPolicySeen" (or other name) is not yet saved in the localStorage.

TODO:
-implement the new slide-in with a conditional (conditions: show if imprint.html present in skin and localStorage flag is not yet saved)
-implement the save and retrieval in the local storage
-link the imprint.html within the text of the slide-in as a link

quasarchimaere:
- Adds a slide-in if a webapp visitor/user has not accepted the disclaimer yet.
- It shows a message with a link within the text (links to the about page) and a button to accept the disclaimer.
-The Status is currently saved into the localStorage and is retrieved on page load and saved in the redux state (subelement of user), accepting also saves the status to the localstorage and the state.

There are 3 new debug methods to be used for testing (these only change the localstorage state and not the redux state!):
```
won.isDisclaimerAccepted() //to see if the disclaimer has already been accepted
won.clearDisclaimerAccepted() //to clear the disclaimer acceptance within the localstorage (refresh should show the disclaimer again)
won.setDisclaimerAccepted() //to programmatically set the disclaimer as accepted (refresh should show no disclaimer then)
```

Please Note that any Text within the imprint.html (thats the thing we will adapt i assume) or the disclaimer itself (including button labels etc) are not in the scope of this PR as this is not my concern (@fkleedorfer will update the texts accordingly)
@fkleedorfer, if there is a need to include more templates into the about page (e.g. if we do not want to use the imprint.html template for everything, please create an issue for that (child of cogsci #1866))
if we do want to update the about page as well then let us push the issue into the ready column in waffle and also make it a child of cogsci (about update issue number is: #1669)

fixes #1888

### DIFF
--- a/webofneeds/won-owner-webapp/README.md
+++ b/webofneeds/won-owner-webapp/README.md
@@ -213,7 +213,10 @@ $ngRedux.getState();
    prevState: {...}
  },
  toasts: {...},
- user: {...},
+ user: {
+    ...,
+    acceptedDisclaimer: true|false, //flag whether the user has accepted the ToS etc. already. (default is false)
+ },
  showRdf: true|false, //flag that is true if rawData mode is on (enables rdf view and rdf links) (default is false)
  showClosedNeeds: true|false, //flag whether the drawer of the closedNeeds is open or closed (default is false)
  showMainMenu: true|false, //flag whether the mainmenu dropdown is open or closed (default is false)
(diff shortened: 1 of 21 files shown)


quasarchimaere:
i added a commit that changes the link so it scrolls directly to the imprint anchor of the about page

peacememories:
The privacy policy page looks a bit glitchy:
image

Issue #1889, opened 2018-06-08
Add Detail for Matching Context


fkleedorfer:
re-enable matching contexts with an optional detail editor.

child of #1866

quasarchimaere:
please be aware that we currently have not implemented this within the detail (because details are only available for is and seeks graphs, the context however is a "toplevel" element of the need

PR #1923, opened 2018-06-13
Re-enable adding matching contexts

sigpie:
Re-enables the adding of matching contexts. Also refactors matching context picking into a separate component. 

Solves #1889


peacememories:
When typing a custom context and clicking the red "x" the default contexts also get cleared, but the checkboxes are not unchecked.

PR #1946, opened 2018-06-14
Fix disable publish on disconnect #1926

### LINKED ISSUE #1926: Publishing needs is not disabled while disconnected
When the connection to the server is lost, it's still possible to create new needs. Those needs only exist as placeholders until the page is reloaded, at which point they'll disappear. Maybe we should disable the Publish button until the session is reconnected.

quasarchimaere:
fixes #1926 

we disable the publish button in the create post view if the connection has been lost or is currently reconnecting

addition:
i also added the same handler for the chat-textfield send (included the accept and send request fields as well)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/chat-textfield-simple.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/chat-textfield-simple.js
@@ -6,12 +6,12 @@
  * Created by ksinger on 16.02.2018.
  */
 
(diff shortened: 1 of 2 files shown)


quasarchimaere:
this pull request has some issue, a reload does not reset the reconnecting ngredux state and thus there is always a "messages" "reconnecting" set to true, which sets the message-send button as well as the publish button disabled, therefore this PR can't be merged as is

peacememories:
I will move the corresponding issue to "In Progress" again if that's alright.

PR #1948, opened 2018-06-14
filter connections (including indicators) based on the state of the r…

### LINKED ISSUE #1739: connections with inactive posts in whatsaround needs
it seems to me that connectionsof whats around needs can contain closed posts (this is probably due to the fact that the whatsaround need is not represented in a connection of the counterpart need and therefore will not get a notification that the need is setto inactive...)

however if its not possible to inform a whatsaround need aboutthe status changes of the connected needs we could at least filter the connections of whatsaround needs if the remoteneed is inactive.

### LINKED ISSUE #1903: own match not filtered after being archived
My own need shows up as a match as soon as I archive it.

Steps to reproduce:
1. create 2 needs that would match
2. archive one of them
3. Reload

Observe that the archived need is now a match to the other one.

quasarchimaere:
…emoteNeed (only use the ones that are actually already in the state and still active)

fixes #1739 
fixes #1903

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-indicators.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-indicators.js
@@ -7,7 +7,7 @@ import won from "../won-es6.js";
 import "ng-redux";
 import { labels } from "../won-label-utils.js";
 import { actionCreators } from "../actions/actions.js";
-import { selectAllOwnNeeds } from "../selectors.js";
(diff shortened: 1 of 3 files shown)


quasarchimaere:
can be tested but must be pushed to int-test first, anyway this will also prohibit the visibility of connections to needs that haven't been loaded yet (or are not reachable) -> we will only show connections that have a corresponding need in the state active (and that is only the case if we were able to fetch the need), this also goes for the counters.

this also fixes the weird ui glitch that shows a connection item without any content but the connectionstate

PR #1959, opened 2018-06-15
Feat  displaying raw trig in needs

m0ru:
When "show RDF" is activated, the entire need RDF is displayed in post-info in the form of TRIG/TTL. I've added a component for this, that the connection-messages now also use.

Caveat: The RDF is completely unfiltered. We might want to limit this to the is/seeks branch or just the `.../event/...#need>`-graph (both of which would exclude creation date and need state)

Depends on #1958


m0ru:
Screenshot with need-RDF displayed

m0ru:
You can use this piece of TTL to test the "arbitrary RDF"-part:
```ttl
@prefix c: <http://example.org/cartoons#>.
c:Tom a c:Cat.
c:Jerry a c:Mouse;
        c:smarterThan c:Tom.

<this:messageuri> won:hasTextMessage "injection working";
                  won:hasTextMessage "injection working2".
```

PR #1960, opened 2018-06-18
Fix disable publish #1926

### LINKED ISSUE #1926: Publishing needs is not disabled while disconnected

quasarchimaere:
fixes #1926 

this also removes the "race condition" that we have had on a reload of the page -> somehow messages.reconnecting within the state is set to true after a reload, however that was never set back to false, now we set both connLost indicator flags to false after the initialLoad has been finished


quasarchimaere:
@maxstolze can we make a separate issue out of this? so we can have a smaller pr for this specific thing

maxstolze:
Outsourced to issue #1987

PR #1983, opened 2018-06-19
Using standard javascript errors.

m0ru:
`Exception` is a custom class defined by the rdfstore, whereas `Error` is the standard class used by the rest of the javascript world. This PR changes all occurrences of the former outside of the store for the latter, which gives us much better support in devtools (e.g. working error messages and stack traces for one)

Resolves #1983

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
@@ -941,7 +941,7 @@ won.toWonMessage = function(message) {
   } else if (message instanceof WonMessage) {
     return Promise.resolve(message);
   } else {
-    throw new Exception(
+    throw new Error(
       "Couldn't convert the following to a WonMessage: ",
       message
     );
--- a/webofneeds/won-owner-webapp/src/main/webapp/.eslintrc.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/.eslintrc.js
@@ -22,7 +22,4 @@ module.exports = {
       },
     ],
   },
-  globals: {
-    Exception: true,
-  },
 };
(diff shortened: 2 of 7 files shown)


quasarchimaere:
this branch seems to be with the old need-reducer.js

webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
```diff
@@ -941,15 +941,15 @@ won.toWonMessage = function(message) {
...
+    throw new Error(
       "Couldn't convert the following to a WonMessage: ",
       message
     );
   }
 };
 
 won.wonMessageFromJsonLd = async function(wonMessageAsJsonLD) {
-  //console.log("converting this JSON-LD to WonMessage", wonMessageAsJsonLD)
+  //console.loErrorting this JSON-LD to WonMessage", wonMessageAsJsonLD)
```
quasarchimaere: seems to be a replace mistake

webofneeds/won-owner-webapp/src/main/webapp/app/reducers/need-reducer.js
```diff
@@ -0,0 +1,1241 @@
+/**
```
quasarchimaere: isnt this the old file? should be removed right?

m0ru:
yeah, i failed to fetch before rebasing ^^"
should be fixed now tho :)

webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
m0ru: should also be fixed now :)

webofneeds/won-owner-webapp/src/main/webapp/app/reducers/need-reducer.js
m0ru: Ah, that was a rebase fuck up. It's fixed now :)

PR #1985, opened 2018-06-19
Not showing private account name in account menu anymore.

### LINKED ISSUE #1979: Use placeholder name for anonymous accounts
Instead of the generated private id, display something like "Anonymous Account" as the user name. Otherwise, people might get confused as to why they appear logged in despite never creating an account.

peacememories:
Fixes #1979

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/account-menu.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/account-menu.js
@@ -12,6 +12,7 @@ import loginFormModule from "./login-form.js";
 import loggedInMenuModule from "./logged-in-menu.js";
 
 import * as srefUtils from "../sref-utils.js";
+import { isPrivateUser } from "../selectors.js";
 
 function genLogoutConf() {
   let template = `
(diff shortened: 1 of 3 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/components/account-menu.js
```diff
@@ -77,7 +77,11 @@ function genLogoutConf() {
     }
 
     getEmail() {
-      return ellipsizeString(this.email, this.maxEmailLength);
+      if (this.email.endsWith("matchat.org")) {
```
quasarchimaere: i think we could use privateIdUser instead (is stored in the state)

peacememories: I looked into the state and tehre doesn't seem to be anything like that?

quasarchimaere: nvm i just saw that i use the routeParam instead...e.g. topnav.js " isPrivateIdUser: !!privateId," should work too though not sure if i like it better than the email ending though...i guess noone will have a user with a matchat.org email anyway... maybe we just need to mark it with a todo for if we ever need to create privateIdUsers from a different domain though

peacememories: I extracted the call into a selector on the state. This should make it more reusable.

peacememories:
In the meantime, can you resolve your change request? :)

quasarchimaere:
sorry... :-(

Issue #1721, opened 2018-05-08
Displaying Needs with Arbitrary RDF


m0ru:
The long-term goal here is to identify and load widgets capable of displaying any RDF to users (see the [RDFruit-community](https://gitter.im/rdfruit/Lobby) for more details on that). In the short-term we need fallback mechanisms. Particular problems/design challenges are:

##  post as list-item in overview

  * title: there might be no title or no text at all
  * placeholder-image: if there's no attached image or title/text our current approach won't work. see  #1720.

## post details

same as above but also: we want to _show something for all triples_. for this we could use `rdfs:label`-schema-entries to render text that's friendlier to non-developers. if we can't display something, we should note it's absence / print an error message and link to the TTL-version.

m0ru:
order of planned implementation:

1. error message for when there's non-displayable rdf
2. `rdfs:label` for key-value style display
3. posts without title =>
  1. update create-post
  2. fallback for title-text
  3. fallback for placeholder-images (see #1720)
4. optional: show-rdf allows displaying rdf in-client instead of linking to server (it's faster this way)

fkleedorfer:
Let's leave 7 for later

quasarchimaere:
i just moved this to in progress, because i think its currently in progress, if this is not the case please forgive me @m0ru

m0ru:
Tinkering with the title placeholder (see screenshots below). Ideally, we'd have smarter ways of deriving "title"-strings for different needs (e.g. "cab ride from $x to $y" for the transport use-case). a somewhat more generic approach would be to look for `xsd:string`s and use the `rdfs:label` linked to their predicates.

end of uri (con: might not work for different nodes in the future)

![end of uri](https://user-images.githubusercontent.com/1730492/41613721-adae68be-73f6-11e8-972b-cb3cfb43f1ae.png)

entire uri (con: very long and thus requires word-wrap; only small difference between lines):

![entire uri](https://user-images.githubusercontent.com/1730492/41613781-dd48b0a2-73f6-11e8-8c9d-44cac1ec844a.png)

generic placeholder (pro: less intimidating; con: every need looks the same but for the identicon):

!["<no title>" as placeholder](https://user-images.githubusercontent.com/1730492/41613871-2a1a5520-73f7-11e8-96c8-9b78f4866a86.png)

fkleedorfer:
[...]

PR #1993, opened 2018-06-20
Feat  optional title

m0ru:
1. create post, but don't enter a title. instead enter abitrary rdf (e.g. `<_:isNeedContent> dc:description "hello world!".`)
2. copy the share-link for the need and open it in another browser
3. make an ad hoc connection.
4. neither the original need nor the ad-hoc need should have a title. the entire view should look like this:


m0ru:
Resolves #1721

m0ru:
You can also close the branch `feat__displaying-arbitrary-rdf-needs` after merging :)

PR #2007, opened 2018-06-22
Fixed #2006 - AnalyzeBehaviour::getUniqueGoalId

### LINKED ISSUE #2006: AnalyzeBehaviour::getUniqueGoalId not unique
AnalyzeBehaviour::getUniqueGoalId returns different strings for the same goal.

Replication steps:
 1. Create a need wich gets matched with the PaypalBot
 2. Type some RDF content into the connection with the PaypalBot which does not fully satisfy the payment goal
 3. Type some other RDF content into the same connection of 2 which completes the payment goal
 4. It will be handled as different goal -> none of those 2 goals are satisfied

schokobaer:
Fixed #2006 

Take the Shapes graph and the data graph and create Strings of those statements. Order them in Array and calculate the SHA-256 Hash of this sorted array. Convert the hash to base64 and return it.


fkleedorfer:
closing in favor of #2008

Issue #2026, opened 2018-06-26
Interface representation of Use Cases


sigpie:
Depends on #1708, #2022 
Related to #1955, #1975, #2024

Use case selection needs a new interface component to select use cases. The list of use cases to display should be read from the config file. This requires refactoring of `create-post` and `create-isseeks`.

Issue #2024, opened 2018-06-26
Display arbitrary details for need creation


sigpie:
Refactor `create-isseeks` to get a predefined list of details and empty draft from the config file. 

Depends on #2021, 
Depends on #2022, 
Depends on #2023
Related to #1955

- [x]  adjust callback methods and template not to be detail-specific
- [x] take a list of details to be displayed as a parameter

PR #2035, opened 2018-06-29
Show arbitrary details in create view

### LINKED ISSUE #1955: Use matching context picker in create-post
Currently, we use the `matching-context-picker` component to tune matching in the `create-isseeks` component, which is used to build the `is` and `seeks` parts of each need. Since the matching context is a need-wide detail, the `need-reducer` gets the matching context from the `is` and `seeks` parts and adds it on the level above. 

At some point, we should refactor this to call the `matching-context-picker` component from the `create-post` component and add the matching context in the correct position instead of adjusting it later. This would also make it easier to refactor detail picking within `create-isseeks` in the future.

### LINKED ISSUE #2010: Move searchString from seeks to root level
We put the `won:hasSearchString` property into the seeks part because that was the easy thing to do. After #1975 is done, we can easily refactor that part to put it into the root level.

depends on #1975

sigpie:
This refactors `create-isseeks` quite a bit to use (almost) no detail names or specifics. While doing this, `matching-context` and `searchString` were also refactored. New posts are now valid as soon as some information is entered. 

Changes in functionality: Posts are valid without `title` or `TTL` as long as at least a single other piece of information is added; `matching-contexts` and `searchString` are now on the same level as `seeks` instead of in `seeks` in both `draft` and RDF; the detail-pickers are currently less pretty - CSS wasn't adjusted as the design will change with #2026  

Fixes #1955 Fixes #2010 
Progresses #2024 (not solved, as detail definitions aren't used yet)


peacememories:
Posting a need without supplying a title or TTL works as expected, but since our master branch is essentially meant to be "live-ready" I would not merge this PR until the style regressions are resolved or the new styles are integrated.

quasarchimaere:
i will merge this PR for now, we will deal with the style regressions in separate issues and prs, as we should not all work on the same branch

PR #2065, opened 2018-07-04
Fix  refresh cncts after reconnect  #2019

### LINKED ISSUE #2019: Refresh connections after reconnect
Currently, when you reconnect after losing the WS connection, the app does not show messages received while the WS was disconnected. Such messages will only become visible once the app is reloaded.

FIx:
* refresh all connections upon reconnect. 
child of #2000

m0ru:
test: 

1. create need
2. copy share link and connect from a different browser-session (accept with the original need)
3. go to the network tab and disable the network connection. then run `ws4dbg.close()` in your terminal (note: you might need to reload before doing that as the dev-tools disconnect doesn't usually apply to web-sockets and `ws.close()` is a bit wonky)
4. write some messages from the other need
5. make another connection request from another browser session
6. set your network to "online" again
7. press "reconnect"
8. the new messages and the incoming request should show up

open issue: the new connection isn't found

Resolves #2019

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/account-actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/account-actions.js
@@ -24,6 +24,8 @@ import { stateGoCurrent } from "./cstm-router-actions.js";
 import { checkAccessToCurrentRoute } from "../configRouting.js";
 
 import { getIn } from "../utils.js";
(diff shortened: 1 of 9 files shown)


fkleedorfer:
Blocked by conflicts, please resolve @m0ru

m0ru:
Should be resolved now :)

PR #2087, opened 2018-07-06
Made labels in custom post more user friendly.

### LINKED ISSUE #2081: Labels of sections in "new custom post"
alternative labels for the sections in new custom post:

* instead of Post: Has/Is
* instead of Search: Want/Looking For

might be clearer than post/search as these sound like document-types/interactions not part of a document, especially given that "post" is also used in the title of the view.

we should also change the placeholders of the text-fields to reflect that.  additionally, we might want to color-code the sections to make them more visually distinct (but that's another issue)

current version, for reference:

image

m0ru:
before: 

image

after:

image

Resolves #2081

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-post.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-post.js
@@ -78,19 +78,21 @@ function genComponentConf() {
         <div class="cp__content">
 
             <!-- ADD TITLE AND DETAILS -->
-            <won-labelled-hr label="::'Post'" class="cp__content__labelledhr" ng-if="self.useCase.isDetails && self.useCase.seeksDetails"> </won-labelled-hr>
+            <won-labelled-hr label="::'Offering / Is'" class="cp__content__labelledhr" ng-if="self.useCase.isDetails && self.useCase.seeksDetails"> </won-labelled-hr>
             <won-create-isseeks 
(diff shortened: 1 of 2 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/components/create-post.js
```diff
@@ -78,27 +78,29 @@ function genComponentConf() {
...
+                title-placeholder="::'What do you want or are you looking for? (Short title shown in lists)'"
                 detail-list="self.useCase.seeksDetails"
                 initial-draft="self.useCase.draft.seeks"
                 on-update="::self.updateDraft(draft, 'seeks')" 
                 on-scroll="::self.scrollToBottom(element)">
             </won-create-isseeks>
 
             <!-- TUNE MATCHING -->
-            <won-labelled-hr label="::'Match'" class="cp__content__labelledhr" ng-if="self.useCase.isDetails && self.useCase.seeksDetails"> </won-labelled-hr>
+            <won-labelled-hr label="::'Matching Behaviour'" class="cp__content__labelledhr" ng-if="self.useCase.isDetails && self.useCase.seeksDetails"> </won-labelled-hr>
```
quasarchimaere: pretty sure we can use || instead of && in thr ng-if otherwise we will only show the labelled-hr before the tune matching behaviour only if both is and seeks are visible, other than that it seems fine to me

PR #2088, opened 2018-07-06
Fix  check for new cncts after reconnect  #2066

### LINKED ISSUE #2066: Look for new connections after reconnect
Similar to #2019 (fixed in #2065) but for new connections (matches and incoming requests).

reproducing the bug:

1. create need
2. copy share link and connect from a different browser-session (accept with the original need)
3. go to the network tab and disable the network connect. then run `ws4dbg.close()` in your terminal
4. write some messages from the other need
5. make another connection request from another browser session
6. set your network to "online" again
7. press "reconnect"
8. the new messages and the incoming request should show up

open issue: the new connection isn't found

child of #2000

m0ru:
## Before

image

## After

image

Resolves #2066

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/account-actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/account-actions.js
@@ -24,6 +24,8 @@ import { stateGoCurrent } from "./cstm-router-actions.js";
 import { checkAccessToCurrentRoute } from "../configRouting.js";
 
 import { getIn } from "../utils.js";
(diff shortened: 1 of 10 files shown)


quasarchimaere:
just my opinion but can we remove the < > i think it makes it look very weird and technical which we dont want, other than that the grey seems to look better

quasarchimaere:
i am also not sure where the changes in the code are that are represented in the screen because the labels that i used are in the labels and afaik the display of the should be in message-content.js so i am not sure where the changes are

PR #2091, opened 2018-07-09
Feat flags #2045

### LINKED ISSUE #2045: "No matches"-flags, etc, are invisible to user
Afaict the user can't notice if a post doesn't get matched or doesn't get matches (e.g. because one of the "don't match" flags are set or the post's "match until" date is in the past). This might lead to a lack of matches/visibility to other people, without users knowing why or being able to find out, which is bad. There should be no hidden system state.

### LINKED ISSUE #2095: Details of matches are not shown
seen on PR #2091

quasarchimaere:
show the flags of a need in the post-content-general section/component

fixes #2045 
fixes #2095


fkleedorfer:
blocked by #2095

PR #2130, opened 2018-07-11
Only produce match if matching context found.

### LINKED ISSUE #2100: pure search finds posts outside of matching context
Steps to reproduce:
1. post a need with a matching context
2. add a searchstring that will find posts without that matching context
Observe that you'll get matches to these posts

peacememories:
Fixes  #2100

### DIFF
--- a/webofneeds/conf/matcher-sparql/matcher-sparql.properties
+++ b/webofneeds/conf/matcher-sparql/matcher-sparql.properties
@@ -14,6 +14,15 @@
 #    limitations under the License.
 #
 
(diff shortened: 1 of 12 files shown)


fkleedorfer:
On integrationtest, this does not change the matching behaviour. Tried with one lunch need in matching context 'SAT', another one without that. the 'SAT' one seems to get a hint from the sparql matcher (score: 1) the other one gets a hint from the solr matcher, it seems (score: 1.422875e-03). The second hint is expected, but the first one should not happen.

PR #2145, opened 2018-07-13
Update tag extraction in utils.js

sigpie:
Tags can be separated by either "," or whitespace now. Only affects the tags-detail-picker.


peacememories:
There is an issue with stray empty tags:

image

PR #2147, opened 2018-07-16
Feat number picker #2142

### LINKED ISSUE #2142: Numbers picker
create a new detail picker for numerical values

quasarchimaere:
Implements an abstractDetail for a number picker,
already use that abstractDetail for the real-estate usecases

fixes #2142


sigpie:
The number picker input field resets if certain characters are entered. Most non-numbers can't be entered, but on entering "e", "-", "+", ".." and possibly others, the input field resets and the previously entered number is deleted.

PR #2204, opened 2018-07-24
Feat newpickertypes #2181

### LINKED ISSUE #2181: Write range pickers for sparql queries
For querying numeric ranges, we need these input widgets:
- [ ] range picker with optional min and max settings
- [x] range picker with preset ranges in a dropdown

quasarchimaere:
fixes #2181 (preliminary work until so we can create the sparql query within the usecase definition)

two new abstract pickers/types:
range-picker -> min-max number picker
select-picker -> either with multiselect (checkbox) or single select (radio)

these should be testable within the testpicker usecase (which we can remove after this is done)


quasarchimaere:
blocked until we chagne the ng-keyup ng-change to a general input event on these fields (i will also change that for all the other pickers too)

peacememories:
^ will be done in a separate pull request

PR #2205, opened 2018-07-24
fixed #2199 (show close instead of backarrow on desktop view for usec…

### LINKED ISSUE #2199: > < in use case headers
This was already fixed for the use case picker in #2158, but is still present for use cases:

arrows

This only occurs on desktop, as the second arrow disappears in responsive mode.

quasarchimaere:
…ase creation)

fixes #2199


peacememories:
It seems the close icons are not perfectly aligned between the usecase view and the details view, but I will merge this anyway.

PR #2208, opened 2018-07-24
Replaced keyup listeners with won-input

### LINKED ISSUE #2206: Change pickers to use input event instead of keyUp
Especially for rich input types the current method of listening to `keyUp` and `blur` is not enough to capture all interactions - e.g. number inputs which have arrows to increment and decrement their value.
Instead we should use the [`input`](https://developer.mozilla.org/en-US/docs/Web/Events/input) event. This is not accessible directly via Angular without using `ng-model`, but we can manually bind the event listener to the DOM node.

peacememories:
Fixes #2206


fkleedorfer:
Testing = checking that everything works as before?

PR #2226, opened 2018-07-27
Feat mandatory details

quasarchimaere:
Created the possibility to set a detail to be mandatory (just set the mandatory parameter to true)

What has been done:
- styled the mandatory details in the create-isseeks view
- added the check for mandatory details in create-post
- altered the existing isValid method by adding the check for mandatoryDetails
- refactored the ols isOrSeeksPresent to isBranchContentPresent, and removed the "fixed" title check, since this is not needed anymore
 
So to recap: 
The Post is valid if ->
- all the mandatoryDetails of both is and seeks are set
- at least one Detail is set within is or seeks (this could also be the case if there are default values set within the useCase itself) e.g. a usecase with fixed/non-editable tags in is or seeks would be valid even if all the visible detailpickers are empty

How To Test:
1) Select the "Rent a place out" usecase
2) see that location and price are mandatory fields, and the submit button is disabled
3) add the mandatory values, and see that publish is possible

mandatorydetail
mandatorydetail2

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-isseeks.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/create-isseeks.js
@@ -24,19 +24,24 @@ function genComponentConf() {
           <div class="cis__detail__items__item__header"
               ng-class="{
                 'cis__detail__items__item__header--won-showvalue': self.details.has(detail.identifier) && self.openDetail !== detail.identifier,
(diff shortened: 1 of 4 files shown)


fkleedorfer:
I dont' see any of that when running locally, using Firefox or Chrome. I can create a realestate need without any data (just the default is and seeks tags)

fkleedorfer:
Sorry, misread the test case. In the realestate offer, there are mandatory fields now.

PR #2227, opened 2018-07-27
fixes #2225

### LINKED ISSUE #2225: realestate search not possible
... if using non-range inputs (no of rooms or size) 

Publishing such a need leads to a js exception (on master)

quasarchimaere:
fixes #2225 

also fixes the problem that if either min or max is not present the post-info would show a "Map{type=xsd:float}" string instead of hiding the element itself

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/config/usecase-definitions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/config/usecase-definitions.js
@@ -863,29 +863,27 @@ const realEstateUseCases = {
           }
         }
 
-        if (filterStrings.length > 0) {
-          const prefixes = `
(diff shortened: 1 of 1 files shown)


fkleedorfer:
The js error is gone, but neither floor size nor room count appears in any query. If you don't select a price range, the query is just 'select ?result'.

quasarchimaere:
there is nothing for roomcount or floor size within the query yet, yes.
however there should not be a select ?result in the query -> i will look into that

quasarchimaere:
app. i was not aware that if we do use the generateQuery function we are not supposed to always return a valid query, so instead of returning undefined (which would have turned into "select ?result" in the hasQuery, i just return our query without any filter arguments

PR #2230, opened 2018-07-27
Added shacl-constraint to range slider for number of rooms

peacememories:
The constraint is not honored anywhere and is not entirely correctly embedded in the rdf (it should be referenced by a `sh:NodeShape`, but it is a valid way to display a range of values for a field that does not have a range as part of its ontology.

As a side effect of `sh:property` possibly appearing multiple times in a need, the need-builder has been updated to merge jsonld branches instead of discarding one of them.

Using jsonld as a working dataset doesn't seem ideal though. Selections on the structure need to be made manually on Immutable.js structures, and prefixed IRIs are not isomorphic to their unprefixed versions.


fkleedorfer:
Please have a look at this: https://github.com/researchstudio-sat/webofneeds/blob/master/documentation/designissues/draft-stating-information-requirements.md

While not perfect and certainly not finished, it's woth revisiting

fkleedorfer:
Goals go in the same direction: https://github.com/researchstudio-sat/webofneeds/blob/master/documentation/designissues/goals.md

PR #2234, opened 2018-07-30
Feat morerangepickers

### LINKED ISSUE #2182: Use Sparql in Real Estate Use Case Def
use the newly created sparql-in-the-client feature to tweak the search case

Requires #2170
Requires #2181

### LINKED ISSUE #2233: add range picker for size in m² for realestate usecase

### LINKED ISSUE #2235: ParseToRDF create-need merge does not work
If you create a need (real estate search) and add a range for floorsize as well as a range for numberOfRooms, there will is only one constraint saved within the need rdf -> See PR #2234 and following rdf https://nodeblue.master.matchat.org/won/page/need/osjlc8zk3dp7

quasarchimaere:
created another rangepicker, and fixed the inclusion of the rangepicker for the number Of Rooms within the real estate offer

however the additon of another picker that produces shackle rdf data seems to fail when there is already one present, e.g one could only add either the number of rooms range or the size range... @peacememories maybe you can see where i am making the mistake

fixes #2233 
fixes #2235 parseToRdf does not use more than one range if defined within a shaclprop... -> see commit 7b191b3 for the solution, the problem was that the array was concatinated but not saved within the original array

fixes #2182 -> the generated sparql query inlcudes the min/max numberOfRooms, min/max Price, min/max floorSize within its filters of a real estate seeks,  it does not include the amenities yet but this is negligible for now

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/details/picker/date-picker.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/details/picker/date-picker.js
@@ -17,7 +17,8 @@ function genComponentConf() {
               class="datep__input__inner won-txt"
               placeholder="{{self.detail.placeholder}}"
               ng-model="self.value"
-              ng-change="::self.updateDate()"/>
+              ng-change="::self.updateDate()"
+              ng-class="{'datep__input__inner--withreset' : self.showResetButton}"/>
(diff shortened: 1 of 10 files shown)


quasarchimaere:
blocked by @peacememories guidance :-)

PR #2240, opened 2018-08-01
Feat  type in use cases

### LINKED ISSUE #2084: set RDF in the usecase definition
We need an option to define rdf:type triples for is and seeks

m0ru:
Look at `usecase-definitions.js:855`, that's how the types are declared. Atm this is demoed with the "search rent" use-case.

Create a need using that use-case. It should have an `rdf:type`/`@type`/`a` triple in it's is- and seeks-nodes.

Resolves #2084


peacememories:
@moru please fix merge conflicts

PR #2246, opened 2018-08-01
Updated Blazegraph and added geolocation queries to Real Estate

### LINKED ISSUE #2191: Honor  Geolocation in Sparql-based matchers
This would have to be done using a sparql extension: https://wiki.blazegraph.com/wiki/index.php/GeoSpatial

peacememories:
Resolves #2191 

Can be tested on Integration Test as of Wednesday 18:30

The radius in which places are matched is hardcoded to 10 km right now.

## Proposed Testing method

1. Create three rentable places in Vienna, Innere Stadt and one in Linz
2. With a new account, create a search in Brigittenau or somewhere close to Innere Stadt
3. The place in Linz should not be matched and, if your search is far enough away, some of the viennese places should also not be matched.

*note*: check for the matcher score. only matches with 100% score are from the sparql matcher. Alternatively, shut down the solr matcher for the test.


m0ru:
It's matching a flat in london with the query for brigittenau

image

peacememories:
This is because for-rent needs don't have a custom query yet, so when you want to rent out a place you'll always match everything with a seeks rent tag

peacememories:
Should we create a new query in this PR or create a new issue for that?

PR #2248, opened 2018-08-02
Feat addcontent message

### LINKED ISSUE #1262: Image-Picker Detail/Viewer

### LINKED ISSUE #2245: implement add attachment for messages
we want 3 options:
[ ] add a workflow
[ ] add an image
[ ] add a file

each option uses different RDF triples to link to the attachment.

quasarchimaere:
This PR implements the ability to add (any detail) content, to a (or as a) chat-message -> this is preliminary work for the process and imagemessages, and uses the same pickers as we do for the need creation.

fixes #1262 
fixes #2245 

The current implementation does NOT use the hasAttachment for the base 64 data, but rather sticks to our current detail/usecase implementation, the file/image detail rdfs are just stubs that might need changing later but it should work for now, one can send any detail as a message now -> however some retrieval functions might not return the expected value due to unknown reasons, the detail is present in the sent and received rdf though, let us fix or refactor the rest in another pull request

addcontent1
addcontent2


quasarchimaere:
do not merge this one yet

quasarchimaere:
This pull is ready to be merged, and is already on integration-test as of 07.08.2018-09:30

PR #2261, opened 2018-08-08
Added feedback to dnd, fixed drop area bug and refactored style

peacememories:


sigpie:
File picker does not accept all file types, e.g. *.md, *.rdf, *.owl and *.klc didn't work. All of them have mime types of text/plain or text/xml.

peacememories:
It seems this problem is Windows-specific. On other systems the OS gives the browser enough information to provide a mime-type, on Windows this is not the case, since it is very file-extension-dependent.
Unfortunately, at the time drop compatibility is checked the extensions of the files are not accessible.
I would suggest we change the accepted mime type field for general files from `"*/*"` to `""` and leave it at that.

PR #2306, opened 2018-08-27
new quick petrinetviewer/picker (similar to file-picker)

### LINKED ISSUE #2294: PetriNet-Workflow Picker/Viewer
Adapt the workflow picker from #2293 to also include the possibility of uploading and displaying petri-net files

child of #2242

quasarchimaere:
fixes #2294 

instead of updating the workflow picker, i have decided to generate a new detail for the upload of petrinets, this is just a quick implementation without any failsafes for correctness of the uploaded (petrinet)-file, it works just like the normal file-picker detail but only lets you upload one specific file.
this should be enough for us to implement the backend side of the workflow process, anything else can or should be dealt with in the future


m0ru:
he meant "one single file"

quasarchimaere:
specific as in "the specific file that the user chose to specify" ;-)

m0ru:
mhm :)

Issue #2342, opened 2018-09-06
Refactor Sparql Matching


maxstolze:
After processing the defaultQuery and post-filtering we need to run a second query to get only the needs with "noHintForCounterpart". For this we need to add a { ?need won:hasFlag won:NoHintForCounterpart } to the query for the second call.
-> SparqlMatcherActor.java:382

And also post filter the results like the first call.
-> SparqlMatcherActor.java:382

maxstolze:
So far the same query is executed and the results are only post-filtered and added to the bulkHintEvent if they are not already in there

maxstolze:
The issue was solved in a different way -> #2350

PR #2343, opened 2018-09-06
Refac sparql matcher behavior

maxstolze:
Blocked by #2342 

The Problem is not solved yet as described in #2342


fkleedorfer:
Found a different solution

PR #2363, opened 2018-09-11
Fix #2362

### LINKED ISSUE #2362: BlankNodes cause duplicates when retrieving RDFStore data
If we have a detail within a connection message that creates blanknodes within the parseToRDF method, a reload of these messages can cause the content to be duplicated (sometimes even tripled, depending on raceconditions), this is visilbe for "images/files" detail for, but also fro the travelAction -> but only the images/files duplication will be visible in the gui, the travelAction and possibly other/new details might not show any visible dupliction other than parseErrors in the console output

quasarchimaere:
updated unittests to represent the updated taxibot goal retrieval

fixes #2362 -> the problem was a racecondition that caused the blanknodes to be stored in the rdfstore multiple times, resulting in contentduplication in chatmessages:
How To Test:
1) Connect with debug bot
2) Send a Location Message
3) Send a TravelAction Message
4) Send a File Message
5) Send a Image Message
6) Send a PersonDetail (including a company)
7) Send a Price
8) Send a PriceRange
9) notice that all the details are visible and not duplicated or missing
10) Reload the page (notice that the messages still did not disappear or duplicate their content) -> if you want to see what the actual problem was just go on a instance without this fix and send a single image as a chatmessage, and reload, you will see that the image is going to be present twice, same with locaiton or travelaction only that a reload will make the location/travelaction invisible due to parsing issues

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/service/need-builder.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/service/need-builder.js
@@ -152,6 +152,7 @@ import { Generator } from "sparqljs";
           ...detail.parseToRDF({
             value: isOrSeeksData[detail.identifier],
             identifier: detail.identifier,
+            contentUri: isOrSeeksData["publishedContentUri"],
           }),
         };
         // add to content node
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/won-message-utils.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/won-message-utils.js
@@ -274,6 +274,7 @@ export function buildChatMessage({
               detail.parseToRDF({
                 value: value,
                 identifier: detail.identifier,
+                contentUri: eventUri,
               });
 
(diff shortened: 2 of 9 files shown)


fkleedorfer:
Does not build on integrationtest (locally it does, though):
```
java.lang.AssertionError: expected:<null> but was:<won.protocol.model.Coordinate@24100000>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at won.utils.goals.instantiation.GoalInstantiationTest.exampleTaxi_validity(GoalInstantiationTest.java:240)
```

quasarchimaere:
fixed the unit-tests this should build now, the problem was that the query did not retrieve all the necessary data and the extractionquery did not return any results -> this was due to the missing name of the to and from Location in the testfile

PR #2368, opened 2018-09-14
Improve message validation

fkleedorfer:
* Revert previous change that tries to ignore message content
* Add checks for properties from the msg namespace - there must not be
any in the message content


fkleedorfer:
works on integrationtest

PR #2400, opened 2018-09-28
Cleaned up details and usecase modularization a bit

m0ru:
Merge #2398 first!


m0ru:
found some issus. don't merge for now

PR #2417, opened 2018-10-03
Jobsearch Use-Case

m0ru:
You can now enter data in "professional » search job", that's tailored to be matched with the hokify job-offers.

Doesn't include yet:

* icons
* sparql-queries
* manually entered job-offers

test:

1. post job search with all details
2. compare the generated RDF against the following json-ld pseudo-code (taken from our modelling document):

```
```json
{
  "won:is" : {
    "@type": "s:Person",
  }

  "won:seeks": {
    "@type": "s:JobPosting",

    "s:employmentType": "Full-time", // full-time/vollzeit, part-time,... - use a dropdown
    "s:industry": ["Computer Software", "Design"], // match by checking for intersection + text-similarity and translations

    //"s:baseSalary": {...}, // free-form-text in hokify json; can't match :|
    
    "s:hiringOrganization": { // for ppl who only want offers from a specific organization (rather niche tho)
      "@type": "s:Organization",
      "s:name": "Google", // JSON - company
    },
 
    "s:jobLocation": [
      {
        "@type": "s:Place",
        "s:geo": {
          "@id": "https://satvm05.researchstudio.at/won/resource/need/rsiiungn085u/location/6l7plycz2g", // unique id; i assume it's necessary for the geo-service
          "@type": "s:GeoCoordinates",
          "s:latitude": "48.216931",
          "s:longitude": "16.361197"
        },
        "s:address": { // couldn't match that atm either + using nominatim guarantees that addresses are resolvable
          "@type": "s:PostalAddress",
          "s:addressCountry": "AT",
          "s:addressLocality": "Vienna"
        }
      },
      {
        "@type": "s:Place",
        //...
        "s:address": {
          //...
          "s:addressLocality": "Graz"
        }
      }
    ],
    
  }
}
```

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/config/details/location.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/config/details/location.js
@@ -1,4 +1,9 @@
-import { getIn, generateIdString } from "../../app/utils.js";
+import {
+  get,
+  getIn,
+  generateIdString,
+  getFromJsonLd,
+} from "../../app/utils.js";
 import Immutable from "immutable";
 import won from "../../app/won-es6.js";
 
@@ -29,11 +34,12 @@ export const location = {
 
 export const jobLocation = {
   identifier: "jobLocation",
-  label: "Jobs Near…",
-  placeholder: "Location: Jobs in Vicinity of...",
+  label: "Job Location",
+  placeholder: "Search for location",
   icon: "#ico36_detail_location",
   component: "won-location-picker",
   viewerComponent: "won-location-viewer",
+  messageEnabled: true,
   parseToRDF: function({ value, identifier, contentUri }) {
     return {
       "s:jobLocation": genSPlace({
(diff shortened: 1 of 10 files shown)


maxstolze:
An error occurs when adding a location to a new custom post and sometimes also to a new job search:

location_bug

A parsing error occurs in the transport route picker:

transport_bug

So something went wrong with the refactoring of location.js

webofneeds/won-owner-webapp/src/main/webapp/config/details/location.js
```diff
@@ -422,3 +209,171 @@ export const travelAction = {
...
+    } else {
+      const locationLat = value.lat && value.lat.toFixed(6);
+      const locationLng = value.lng && value.lng.toFixed(6);
+      if (locationLat && locationLng) {
+        humanReadable = "@(" + locationLat + " , " + locationLng + ")";
+      }
+    }
+    if (humanReadable) {
+      return includeLabel
+        ? this.label + ": " + humanReadable.trim()
```
sigpie: This use of `this.label` causes the error documented by @maxstolze - `this` is expected to be the detail definition, which is no longer true due to the refactoring.

m0ru:
Should be fixed now. Please avoid using `this` in javascript, as it's notoriously unclear what it refers to. Rather stick to pure/side-effect free functions where everything is contained in the function signature as far as possible :)

PR #2419, opened 2018-10-04
Fix for the NeedInactivityChecker

fkleedorfer:
Now supports an inactivity interval of 0, which deactivates it.


fkleedorfer:
How to test: set `need.inactivity.check.interval=0` and see if the node starts

PR #2426, opened 2018-10-11
Feat detailstruct review

quasarchimaere:
builds a rateDetail-structure via its own detail-picker/viewer

this is preliminary work for the review/reputation feature that we are going to implement

Please check the picker in general, i think there could be some update issues depending on which element you add/edit first or at least try if a re-edit is going to be processed in the data after publish -> currently you can add the detail in the costum post section as well as in any open chat (i know the rating during need creation does not make alot of sense but we currently dont have the ability to define message only details...)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/details/picker/pickers.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/details/picker/pickers.js
@@ -15,6 +15,7 @@ import priceRangePickerModule from "./price-range-picker.js";
 import filePickerModule from "./file-picker.js";
 import workflowPickerModule from "./workflow-picker.js";
 import petrinetPickerModule from "./petrinet-picker.js";
+import reviewPickerModule from "./review-picker.js";
 
 /**
  * module names for angular's module system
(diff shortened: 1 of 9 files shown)


sigpie:
Currently adding reviews as detail to a conversation causes an exception. This is due to the review value being saved with the type `xsd:in` in review.js:54, but being parsed as a `xsd:string` in review.js:71 

Changing review.js:71 to `xsd:int` causes another exception, as the switch in won-utils.js:500 does not expect the type `xsd:int`.

sigpie:
Viewing a custom need with a review detail also causes an exception, despite the need being displayed correctly: 

review-exception

As this is looking for `https://localhost:8443/owner/{{ self.copied ? '`, it seems as if the wrong url is added at some point

sigpie:
When adding a review to a conversation, the send-button is initially disabled despite a rating being selected: 

disabled-send

The send button becomes enabled if the rating is changed, a the review-textfield is changed or message text is added. If message text is added while the review detail is still pristine, no review information is sent. To prevent confusion, "Select a rating" would probably be a better default value than "3 Stars".

quasarchimaere:
i think the self.copied part could possibly come from the share post link component

sigpie:
> i think the self.copied part could possibly come from the share post link component

It is, see #2427

quasarchimaere:
fixed the parsing error
added "Select Rating"-undefined default value

branch should be mergable now

PR #2437, opened 2018-10-16
Support for creating and using personas

### LINKED ISSUE #2374: Design for Identities
Design the owner gui such that it is possible to
* create multiple identities per user acccount
* link one identity to a need when creating it
* display identities on own needs in list
* display identites on matched needs

peacememories:
Fixes #2374 

This PR adds a persona creation menu accessible from the account menu (`Account Settings`).
It also changes the publish button to allow selecting with which persona to publish a post.

> Note: To test this PR you will need to run `npm install`, as it depends on many new packages
> You will also need to rebuild your owner webapp. As of now the application is also available on the integration-test server

Where to watch for breakage:
* The redux state should correctly show the personas as needs, and correctly show connections between them and posts they are attached to
* The connection overview (left) should not show any personas either as posts or as connections
* The publish button should correctly display all successfully saved personas, but no unsaved ones, and no other needs
* The persona settings should show personas which are currently in the process of being saved as "Saving...", and should not show any posts


quasarchimaere:
please make sure that the selectAllNeeds selector in selectors.js only returns AllPosts and not AllNeeds, otherwise our suggest posting detail will result in letting others share persona needs as well (via the gui), not sure if we want to have this possibility in the current iteration of the persona implementation

sigpie:
The debug bot works for me on integration-test, creating needs with personas work and both personas and needs show only up where they are supposed to. 

The posting detail referenced by @quasarchimaere has been adjusted to work with the changes from this PR.

PR #2452, opened 2018-10-18
Feat claim ui

quasarchimaere:
Implements the action to "claim" a single message as well as a set of messages
The Process in the UI is the exact same as for proposes messages, with the only change that the sent message does not use the agr:proposes but the agr:claims predicate


quasarchimaere:
will be included within another pull

PR #2456, opened 2018-10-19
Made connection list and connection content scale symmetrically.

### LINKED ISSUE #2335: 50:50 column width and break-point
right column almost non-existantly slim

The column should have even width imo and the breakpoint should prevent anything from getting that squeezed.

peacememories:
Fixes #2335 

Setting `flex-basis: 0` on both elements makes them scale exactly equally. This is a fix for the squished right side, but it is not perfect visibility-wise, as seen in this screenshot:
image

To fix this, we need to change how we do responsiveness. A single "breakpoint" isn't going to cut it there.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/style/_connections.scss
+++ b/webofneeds/won-owner-webapp/src/main/webapp/style/_connections.scss
@@ -17,14 +17,13 @@
     flex-grow: 1;
     max-width: $maxContentWidth / 2;
 
-    @media (min-width: $responsivenessBreakPoint) {
-      min-width: 28rem;
-    }
+    flex-basis: 0;
   }
 
   .conn__content {
     display: flex;
     flex-grow: 1;
+    flex-basis: 0;
     height: 100%;
     justify-content: center;
     align-content: center;


quasarchimaere:
i do not like this "fix" at all, we see that the left side was moved in a way that the indicators move over the title, there was a reason why the left side was set to a certain minimal width to begin with, the only problem was that the responsiveness breakpoint was not met yet and thus resulted in the squished right side only... imho i would not merge this PR since it replaces one issue with another

fkleedorfer:
Can't we have the original behaviour and add a minimum width for the right side, under which it won't be displayed?

peacememories:
Not exactly. We could add a minimum-width to the second column, which would result in both columns having a minimum width, which would either lead to the same behaviour we originally had, or to the columns spilling over the viewport (I haven't checked which).

Not displaying an item when it reaches a certain threshold is also not possible (*element queries* have so far been rejected as proposal, because they could result in infinite loops in the render logic). We could add another breakpoint specific to this page, but that is exactly my point. We need a different way to handle responsiveness, or we at least need to be aware of and live with the consequences of adding breakpoints "randomly" to the design.

fkleedorfer:
What would you suggest?

quasarchimaere:
#2523 changes the structure of the views, we can't merge this pr anylonger the changes have become obsolete

PR #2459, opened 2018-10-19
Change handling of message coming from the WoN node

fkleedorfer:
Instead of
1. trigger server-side actions
2. process message and forward to client
do
1. process message and forward to client
2. trigger server-side actions

This should reduce the risk of unexpected message order seen by the
client (e.g. receive connect with newly created need before receiving
the success response to the create)


peacememories:
How could we best test this?

fkleedorfer:
this is a change to server-side actions, so it can be tested by using the connect action. Basically posting with persona is the test. If it still works, the PR is good. You may notice that there is less latency before you get messages back than before, because message delivery to the owner app client is now done before handling server-side actions (which may include sending messages to the node again)

PR #2464, opened 2018-10-23
Displaying author and rating on connections

peacememories:
Ratings are currently only mocked up, but the interface is fully functional.

Rating a persona does not do anything, and the displayed rating is derived from the persona uri.


sigpie:
tested by @m0ru

PR #2466, opened 2018-10-24
Feat minify messages

quasarchimaere:
In Order To make our chat-views more useful we try to minify/collapse messages or their respective content so that the view is not cluttered with cascading and duplicate entries

This PR changes the following in the chat-view (aka open connection view)

Any message that has been proposed/claimed/rejected/retracted will be minified in the chat view (see screenshot below
collapsedmessage
)

The referenced-content of a message will be collapsible/expandable now, the references have the following inital collapsed State:
- forwards -> not collapsed by default
- proposes -> not collapsed by default
- claims -> not collapsed by default
- accepts -> collapsed by default
- proposeToCancel -> collapsed by default
- rejects -> collapsed by default
- retracts -> collapsed by default

If a message containing references is going to be collapsed (e.g if the message was rejected) the contained references of the message are also all collapsed -> this will ensure that upon displaying (click on the collapsed message) the message we will not get bombarded with the whole content right away

### DIFF
--- a/webofneeds/won-owner-webapp/README.md
+++ b/webofneeds/won-owner-webapp/README.md
@@ -148,6 +148,15 @@ $ngRedux.getState();
                            isSelected: true|false //whether or not the message is Selected in the MultiSelect view
                            isCollapsed: true|false //default is false, whether or not the message should be displayed in a minimized fashion
                            showActions: true|false //default is false, whether or not the actionButtons of a message are visible
+                           expandedReferences: {
+                             forwards: true,
+                             claims: false,
+                             proposes: false,
+                             proposesToCancel: false,
+                             accepts: false,
+                             rejects: false,
+                             retracts: false,
+                           }
                        },
                        uri: string //unique identifier of this message (same as messageUri)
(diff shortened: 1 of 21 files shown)


fkleedorfer:
It's better now but still inconsistent and hard to follow. Consider a proposal. 
It starts with a message (M) that message gets hidden when it s proposed by proposal P. 
When P gets accepted by A, though, P gets different buttons, but keeps its title.

I think the transition M -> P is good, but P -> A is inconsistent with that and not as good.

How about this:
as P gets accepted by A, P is hidden (just as M was earlier), saying 'Message was accepted. Click to expand', and A starts expanded - same for propose to cancel and accept.
good:
grafik
not so good:
grafik
I find this confusing because the state of the agreement protocol here is already 'agreed', but the prominent interaction element says 'proposes'. Confusingly, the button says 'propose to cancel'. 

# Wording
I find the titles confusing. How about: 
* 'proposing N messages' (works both if I say it and if my counterpart says it)
* 'accepted N messages' (same)
* 'claiming N messages'
* 'canceled N messages'

PR #2485, opened 2018-10-31
Debug output cleanup

sigpie:
Removed/changed most occurrences of `console.log`. Unused output was deleted, everything else was set to `console.debug`, `console.warn` or `console.error`. 

This should reduce the amount of console output while the app is running (console output detail level may need to be adjusted) and not break anything.


sigpie:
Resolved all merge conflicts.

PR #2486, opened 2018-11-01
Fix sparql query transform

fkleedorfer:
The embedded sparql query gets transformed by the matcher before
execution, adding checks for NoHintForCounterpart flags.
The problem is that this change was applied to each subquery instead of
just the toplevel one. When there is a subquery using 'group by' - the
transformation leads to
    ...
    GROUP BY ?result
    HAVING NOT EXISTS {
     ?result ...

Which blazegraph isn't able to execute, in contrast to jena


fkleedorfer:
tested on integrationtest

PR #2488, opened 2018-11-01
Feat musician matching

fkleedorfer:


fkleedorfer:
tested on integrationtest - working.

PR #2496, opened 2018-11-05
Refac selectors

quasarchimaere:
Refactors/Split selectors.js into sep. files, extract pure util functions to separate util files (e.g. isMessageUnread)

Created helper funtions (e.g. isChatConnection -> if a message has the facet ChatFacet)

Renamed all selectors from select... to get...
Removed unused selectors

Implemented chat-indicators so they only take chatConnections into consideration


webofneeds/won-owner-webapp/src/main/webapp/app/selectors/connection-selectors.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Get all the connectionUris storid within the state
+ */
+export function getOwnedConnectionUris(state) {
+  const connections = getOwnedConnections(state);
+  return connections && connections.keySeq().toSet();
+}
+
+export function getChatConnectionsByNeedUri(state, needUri) {
```
peacememories: This function fails because it is called with wrong arguments:

```
TypeError: state.get is not a function
    at getNeeds (general-selectors.js:16)
    at getChatConnectionsByNeedUri (connection-selectors.js:43)
    at Controller.hasOpenOrLoadingChatConnections (connections-overview.js:368)
    at fn (eval at compile (angular.js:15551), <anonymous>:4:683)
    at Scope.$digest (angular.js:18253)
    at Scope.$apply (angular.js:18531)
    at angular.js:20448
    at completeOutstandingRequest (angular.js:6362)
    at angular.js:6642
```

PR #2499, opened 2018-11-06
Fix personas in re posts

### LINKED ISSUE #2471: Post AdHoc with persona
When connecting to a need N via WhatsX, it isn't possible to create the Re:N need with a persona, which is inconsistent with normal need creation.

sigpie:
Closes #2471 

Added persona selections to adhoc connections, i.e. connections started from a WhatsX-Post or from a shared Post-URI. 

Things to check:
* persona selection is only visible if the user is logged in
* persona selection is only visible where it's useful, that is creating a new post, not for any actions that do not create a new post
* the submit button (which is next to every text field) does behave as expected

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/connections-actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/connections-actions.js
@@ -328,11 +328,18 @@ function connectAdHoc(theirNeedUri, textMessage, persona, dispatch, getState) {
         body: JSON.stringify([
           {
             pending: false,
-            facet: `${persona}#holderFacet`,
+            //facet: `${persona}#holderFacet`,
+            facet: getIn(state, ["needs", persona, "hasFacets"]).keyOf(
+              "won:HolderFacet"
+            ),
           },
           {
             pending: true,
             facet: `${needUri}#holdableFacet`,
+            // FIXME: does not work as new need is not in state yet
+            //facet: getIn(state, ["needs", needUri, "hasFacets"]).keyOf(
+            //  "won:HoldableFacet"
+            //),
           },
         ]),
         credentials: "include",
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/connections-actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/connections-actions.js
@@ -311,10 +311,39 @@ function connectAdHoc(theirNeedUri, textMessage, persona, dispatch, getState) {
     const theirNeed = getIn(state, ["needs", theirNeedUri]);
     const adHocDraft = generateResponseNeedTo(theirNeed);
     const nodeUri = getIn(state, ["config", "defaultNodeUri"]);
+
+    // create new need
     const { message, eventUri, needUri } = await buildCreateMessage(
       adHocDraft,
       nodeUri
     );
+
+    // add persona
+    if (persona) {
+      const response = await fetch("rest/action/connect", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify([
+          {
+            pending: false,
+            facet: `${persona}#holderFacet`,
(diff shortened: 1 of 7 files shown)


quasarchimaere:
i find the button label rather long, i know this is our standard label but in this case we should figure out something else because its too long on desktop and def. too long in mobile. so if we think the missing padding is an issue we should see the labellength as an even more severe issue

webofneeds/won-owner-webapp/src/main/webapp/app/actions/connections-actions.js
```diff
@@ -301,20 +301,49 @@ export function connectionsOpen(connectionUri, textMessage) {
...
+    if (persona) {
+      const response = await fetch("rest/action/connect", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify([
+          {
+            pending: false,
+            facet: `${persona}#holderFacet`,
```
quasarchimaere: i think the actual holderFacet or holdableFacet uris are stored within the state... let us not assume from the getgo that the uri will look like this but instead retrieve the uri from the need in the state. hint: need.get("hasFacets") -> retrieves a map with facetUri > facetType so we should  be able to find the correct facetUri if we search for the correct facetType... if this is the case in our other connectAction (the one @peacememories implemented we should change it there aswell)

webofneeds/won-owner-webapp/src/main/webapp/app/components/post-messages.js
```diff
@@ -226,9 +226,10 @@ function genComponentConf() {
 
             <chat-textfield-simple
                 placeholder="::'Message (optional)'"
-                on-submit="::self.sendRequest(value)"
+                on-submit="::self.sendRequest(value, selectedPersona)"
                 allow-details="::false"
                 allow-empty-submit="::true"
+                show-personas="::true"
```
quasarchimaere: shouldnt we only display personas if the user has them?

sigpie: This is just a flag that's used to signify that this button can have a persona selection, the actual selection is only visible if the user is also logged in. 

It would be possible to also hide the selection if the logged-in user has no personas, but I think in this case it would be better to display an add persona button or a message that there are no personas yet.

sigpie:
Padding has been added in all cases now. I agree that the labels are too long, but I couldn't think of a shorter label to use instead. Even if we shorten the "Anonymous" part, we'd still need to add the persona name (regardless of length) after a persona is selected. 

For now, the button label wraps and is displayed in two lines, but this needs some further investigation by @peacememories.

quasarchimaere:
An option could be to keep the label we had before instead of the "...as Anonymous"
If a Persona is selected we could write "...as [Identicon]" (and show the identicon of the persona instead of the name, since the name would be visible in the selection menu anyway

webofneeds/won-owner-webapp/src/main/webapp/app/actions/connections-actions.js
sigpie: This is only possible for the holderFacet. The holdable Facet of the created need can't be accessed via the state, as it is not in the state yet when the persona connect message is sent.

sigpie:
I can see keeping out the "Anonymously" as an option to make the labels shorter. Adding "as [Identicon]" would probably look weird though and might add other layouting issues. Without wrapping the text, it would still be too long for small displays (i.e. more than half the screen width). With wrapping the text, the button would need to be higher, which would also need to increase the height of the text field and the add details button. 

Also, I don't want to use a different format for displaying personas here compared to the publish button at post creation. Two elements with the same functionality should also have the same look.

webofneeds/won-owner-webapp/src/main/webapp/app/components/post-messages.js
quasarchimaere: ok

PR #2500, opened 2018-11-07
Feat paypalpayment detail

quasarchimaere:
Creates a paypal-payment detail for messages, so that the paypal bot can be used more easily (paypal bot adaptations pending)


webofneeds/won-owner-webapp/src/main/webapp/config/details/payment.js
```diff
@@ -0,0 +1,154 @@
...
+          "@id": "gr:PayPal",
+        },
+        "s:paymentStatus": { "@id": "s:PaymentDue" },
+        "s:totalPaymentDue": {
+          "@id": idString ? idString + "#totalPaymentDue" : undefined,
+          "@type": "s:CompoundPriceSpecification",
+          "s:price": [{ "@value": value.amount, "@type": "xsd:float" }],
+          "s:priceCurrency": value.currency,
+        },
+        "s:paymentMethodId": value.secret, //TODO not sure if this would be the correct predicate for our Secret
```
sigpie: [s:identifier](https://schema.org/identifier) might be a better fit than s:paymentMethodId here.

sigpie:
* It's possible to enter a negative price (or a price of 0)
* "Who pays" can't be selected unless all other fields have content
* the receiver id should be checked to be an actual email address
* "Who pays" is an unclear label, is this supposed to be the connection the payment is for? Or is this a selection for who will pay any additional payment fees?
  * If it's supposed to select the connection the payment is for, it should not show owned needs - with owned needs it's confusing, especially if there is only one connection.
  * If it's supposed to specify who will pay additional fees, it would be better to just have a "you"/"them" selection instead of showing specific needs. In this case a need selection is missing.

PR #2517, opened 2018-11-14
Fixes #2516 close conversation button for all connection states

### LINKED ISSUE #2516: Add Close conversation button to conversation context menu
As described in #1810, closing conversations is inconsistent in our current user interface. I suggest that all conversations should have a context menu with a "close/remove conversation" button, regardless of state.

quasarchimaere:
Conversation/Connection Context Dropdown has a "remove connection" button regardless of the connection state:

Here are the displayed labels:
connected -> "Close Connection"
suggested -> "Remove Connection"
SentRequest -> "Cancel Request"
ReceivedRequest -> "Deny Request"

Maybe we should not use a different label since every button results in a closeConnection request, if that should be the case i will adapt this PR accordingly


sigpie:
Does not work as intended:

label

PR #2523, opened 2018-11-19
Feat gdpr aboutpage

### LINKED ISSUE #1669: Update About Page

### LINKED ISSUE #1772: Make about page accessible from anywhere, easily
Some first-time users feel lost on our start page. Let's show them a friendly About link/button.

Suggestion: at the bottom of the page

### LINKED ISSUE #2282: Create and place"Help" or "About" section at hand
I think we need to create a "Help" o "About" section and place it at hand from the very beginning. I like how Pinterest do it. They have to "About" sections: a first shorter one and a second longer one. On the first one there's a quick explanation about Pinterest and an invitation to try it. Here I leave you a suggestion (for brainstorming):

artboard

quasarchimaere:
Refactors the about page according to #1669 and #2282
Implements a footer for connections.html / about.html / settings.html / signup.html / post.html
to be able to access the about page and any other link that we want to have (could also include the fork me @ github)

fixes #1669 
fixes #2282 
fixes #1772


sigpie:
* As mentioned by @m0ru in #1772, putting the about page in the footer makes it harder to access if there are a lot of existing needs. If there are no posts yet, it's easily accessible, but it's hard to get back to the footer on long pages, e.g. the about page. 
* The new tutorial remembers the scrolling position from the main view, which means that on a small display, the top of the tutorial is cut off when opening it.
* Pages linked in the about also remember scrolling positions, which is confusing especially as most of the links are sections of the about page but don't go to the about page - e.g. the privacy policy should be either on the about page and the footer link should open the about page scrolled to the correct position or it should not fully be on the about page and instead open a separate page as the footer-link does now. 
* footer elements should wrap as elements, not within elements

footer_wrapping

I'd also argue for removing the "What's new" and "What's around" buttons from the tutorial. "Post something now!" leads to a view that offers these buttons anyway. Also, they hide the "Post something now!" button from view due to scrolling on small screens.

quasarchimaere:
Here are my answers to the comment made by @sigpie:

> * As mentioned by @m0ru in #1772, putting the about page in the footer makes it harder to access if there are a lot of existing needs. If there are no posts yet, it's easily accessible, but it's hard to get back to the footer on long pages, e.g. the about page.

If there are alot of needs the chance of the User needing the about page is very limited (imho), the footer is not fixed to the bottom as we probably will use it as a "sitemap" like feature with many many links to come. I might also note that the access to the about page is (although not ideal) at least present in every instance whereas before it was not accessible to the user if there was noone signed in yet.

> * The new tutorial remembers the scrolling position from the main view, which means that on a small display, the top of the tutorial is cut off when opening it.
> * Pages linked in the about also remember scrolling positions, which is confusing especially as most of the links are sections of the about page but don't go to the about page - e.g. the privacy policy should be either on the about page and the footer link should open the about page scrolled to the correct position or it should not fully be on the about page and instead open a separate page as the footer-link does now.
The about page scrolling position should be handled in a separate issue (see issue #2527)

> * footer elements should wrap as elements, not within elements
> 
> footer_wrapping

see "note" of PR Description, footer styling is still pending changes and must be adapted along the way in the future


> I'd also argue for removing the "What's new" and "What's around" buttons from the tutorial. "Post something now!" leads to a view that offers these buttons anyway. Also, they hide the "Post something now!" button from view due to scrolling on small screens.

I think the duplication of these elements does not hurt us, so we can probably leave it in there (if not we should create an issue and discuss it)

PR #2544, opened 2018-12-04
Feat verification screens #2538 rebased

### LINKED ISSUE #2538: Implement E-Mail Verification Screens
All The email verification endpoints result in a json response now, that is just visible as plainText in the browser -> we should create a better UX for these endpoints:

either as static html pages or implement it within our own single page app

### LINKED ISSUE #2539: Implement Resend Verification Mail as State Reducer
Currently we only call the resend verification mail ws-method from `login-form.js` and `slide-in.js`

We might want to create a reducer that stores the initial invocation and the response of the ws-call in the state, right now we are not parsing the ws-response at all, we simply check if the button has been clicked and reset this after i think 2seconds again

### LINKED ISSUE #2545: No visible reaction to login failure
Trying to log in with a wrong username/password combination leads to no visible reaction of the GUI. The following error is logged:
FF:
```
Unhandled promise rejection TypeError: "s.actionCreators.account is undefined"
	B https://ownergreen.master.matchat.org/owner/generated/app_jspm.bundle.js:8:221165
	s https://ownergreen.master.matchat.org/owner/generated/app_jspm.bundle.js:10:72416
	Y https://ownergreen.master.matchat.org/owner/generated/app_jspm.bundle.js:10:72561
	u https://ownergreen.master.matchat.org/owner/generated/app_jspm.bundle.js:10:76382
```
Chrome:
```

/owner/rest/users/signin:1 Failed to load resource: the server responded with a status of 403 (Forbidden)
account-actions.js:201 Uncaught (in promise) TypeError: Cannot read property 'loginFailed' of undefined
    at account-actions.js:201
```

### LINKED ISSUE #2546: Existing posts can't be transferred when creating an account
This only occurs on master, not on live. If posts are made with an anonymous user account and the sign-up process is started, the "Keep Postings" action fails. "Start from Scratch" works as intended. 

To reproduce: 
* create a few posts while not logged in
* click "Sign Up"
* enter your email and password
* click "Keep Postings"

### LINKED ISSUE #2547: email verification only works after logout+login
Steps to reproduce:
1. sign up new user
2. observe that there is a slidein asking you to verify your email address
3. wait for verification email
4. click link in email
5. reload owner webapp
6. observe that the slidein is still there
7. logout, login again
8. observe that the slidein is not there any more

Reason is that signin loads the user from the db, isSignedIn retrieves the user from the session. The user object retrieved by isSignedIn is stale and needs to be updated

quasarchimaere:
fixes #2538 
fixes #2539 

fixes #2545
fixes #2546
fixes #2547

Implements the verification-"screens" as a slide-in:

- Changes the verification link in the email to [ownerapp]/connections?token=[token]
- If the slide-in sees that a token is in the url it will check the token on load and display the result of the call in the slide-in

Which slide-in(text) will be displayed and when?
The Verification Slide-In itself will only be displayed if there is no connectionLost slide-in and if there is either a verificationToken in the url or a logged In user that has NON verified email adress

- "E-Mail has not been verified yet, check your Inbox." -> if there is no verificationToken in the url and the email is not verified yet
- "Verifying the E-Mail Address" -> if there is a verificationToken in the Url and the validation is currently being processed
- [EmailVerificationError] (e.g token not found, token expired) -> if the validation returned an error and the validation is not in process and there is a validationToken in the url
- "E-Mail Address verified" - if the user is logged in and there is a verificationToken in the url and there is currently no validationProcess running and there are no Errors
- "E-Mail Address verified (Please Login now) -> same as the above but without a user being logged in, this is necessary because we can't automatically induce a login without a rememberMe token or a password to unlock the keystore

- Additionally a Loading Spinner will be shown if there is currently a validation or a resend(verification mail) in progress
- We also show a close button to close the verification Slide in that is present if the validation is not in process and the validation was successful and there is a verificationToken in the url -> this close button removes the verificationToken parameter from the url (with a stateGoCurrent)
- A "Resend Mail"-Button is shown if neither the verification nor the resend is in progress and if there is either a unverified user signed in or there is a verificationToken present in the url and there was a emailVerificationError)


quasarchimaere:
The verification is not automatically set to true after a reload (when logged In) because the user is retrieved from the auth principal and not from the service, we need to retrieve the User from the userService and genereate the response from that user (see RestUserController.java:346)

sigpie:
> "E-Mail Address verified (Please Login now) -> same as the above but without a user being logged in, this is necessary because we can't automatically induce a login without a rememberMe token or a password to unlock the keystore

Does not work for me. Using a token while not logged in shows no slide in. The address is also not successfully veryfied.

quasarchimaere:
Do not merge yet, we should fix the mail link and the two issues in progress too

quasarchimaere:
I fixed the two issues that were found in the  review of this PR (#2546 #2547) and updated the link in the verification mail.

If the verification Link still gets escaped when opening it then we have an issue with all of our e-mail links and should handle this in a sep. PR and create an issue for it

PR #2567, opened 2018-12-10
Created account export feature

### LINKED ISSUE #1821: Account Export
requires #2521

peacememories:
There are still two unfinished steps:
* Allow anonymous users to export their accounts by providing an email field
* Make the navigation sidebar mobile-friendly

Apart from that, more descriptive error messages should be added to the export result.

Fixes #1821


webofneeds/won-owner-webapp/src/main/java/won/owner/pojo/RestStatusResponse.java
```diff
@@ -33,7 +33,10 @@
 
     SETTINGS_CREATED(5200, "Settings created", HttpStatus.CREATED),
 
-    TOS_ACCEPT_SUCCESS(6200, "Successfully accepted Terms Of Service", HttpStatus.OK);
+    TOS_ACCEPT_SUCCESS(6200, "Successfully accepted Terms Of Service", HttpStatus.OK),
+
+    EXPORT_SUCCESS(7200, "Successfully started exporting user", HttpStatus.OK),
+    EXPORT_IS_ANONYMOUS(7403, "To export with an anonymous user you need to provide an email address", HttpStatus.BAD_REQUEST);
```
quasarchimaere: As far as i see these response codes are not pushed into the`won.js`-ResponseCodes yet, we need to add those before we can merge this PR

webofneeds/won-owner/src/main/java/won/owner/model/User.java
```diff
@@ -144,6 +144,8 @@ public boolean isCredentialsNonExpired() {
         return this.emailVerified || isWithinGracePeriod();
     }
 
+    public boolean isPrivateId() {return this.email.endsWith("@matchat.org");}
```
quasarchimaere: we should change this into a won_owner-db saved attribute instead, need to update the db-schema, can be done within this PR or a sep. PR but has to be done regardless, after we added this attribute we should set the correct value of it while a signup/transfer happens, and push the attribute to the client after login/isLoggedIn as well

quasarchimaere:
maybe not in scope for this pr, but we should think/worry about attachment size limitations, it could easily happen that we are unable to send the email if the export file is too big for the recipients mailbox/server and thus the export will never reach the user(specified email), i would change this in favor of sending a download link in the email. this link could grant access to a (one time?) download of the export, if its accessible more than once we should at least invalidate/delete the entry after a certain date or if the user has requested a new export.the benefit of this approach would be that we could link the export file from within the owner webapp as well (and possibly deny subsequent exports for some time, as this is probably a very costly operation)

webofneeds/won-owner-webapp/src/main/java/won/owner/pojo/RestStatusResponse.java
peacememories: Okay, will do, although js isn't involved in the request here (I found it a bit of a waste to route a simple request through our whole redux system)

webofneeds/won-owner/src/main/java/won/owner/model/User.java
peacememories: I think this should be done in a separate PR, to keep this PR focused on the feature it is trying to add.

peacememories:
We should definitely worry about this, but solving that problem would mean creating a download manager system like google has it for data exports.

webofneeds/won-owner-webapp/src/main/java/won/owner/pojo/RestStatusResponse.java
peacememories: Response codes are now added to `won.js`

PR #2568, opened 2018-12-10
Feat  save anon link #2283

### LINKED ISSUE #2283: Save anonymous users's link
requires #2554

For the users who decided to stay anonymous there's a risk of losing their link with their activity. From a previous prototype I suggested something like this:

link email

quasarchimaere:
Implements the anonymous-warning slide-in (based on issue #2283)

Fixes #2283 

Anon slide-in is currently shown everytime a user is logged in and there is a privateId set in the url -> this should/needs to change once we return isAnonymousAccount after signup/transfer/login, so we can remove the privateId in url handling after all and just use it as a login entry point


quasarchimaere:
Copy link to clipboard doesnt seem to work because of the z-index impl, we might need to use a different approach

PR #2576, opened 2018-12-11
Feat  real anon check

quasarchimaere:
Adds the "real" check for isAnonymous on the Owner backend, and adds frontend handling of anonymous users (-> do not show the emailVerification-Slide-in if an anonymous user is logged in)


quasarchimaere:
This PR is so we can (finally) start implementing handlers to avoid sending anonymous Users emails into the void

PR #2583, opened 2018-12-12
Feat classifieds usecases

### LINKED ISSUE #2512: Add Classified UseCase Group + UseCases
child of #2322 
requires #2520

Implement "Classifieds"-UseCaseGroup

Add UseCases
"Offer Goods"
"Offer Services"

We could also add their Search Counterparts, but for now i would just use the simple search to cover it.

sigpie:
Adds usecases for "classified ads" that allow offering and searching for objects and services. The demand use case has a custom sparql query that matches to offers containing keywords, the offer use cases don't have custom queries yet. Matches should be symmetrical (i.e. hints sent to both needs), but currently aren't, only demands get hints.

This also changes a property in `webofneeds/won-docker/image/bigdata/RWStore.properties#33` that enables text indexing. As this is a potentially expensive operation, this may impact performance.

Closes #2512

### DIFF
--- a/webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
+++ b/webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
@@ -149,6 +149,7 @@ public class SparqlMatcherActor extends UntypedActor {
             log.info(String.format("Caught exception when processing %s event %s. More info on loglevel 'debug'",
                     eventTypeForLogging, uriForLogging.orElse("[no uri available]")));
             log.debug("caught exception", e);
+            e.printStackTrace();
         }
     }
 
--- a/webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
+++ b/webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
@@ -149,7 +149,9 @@ public class SparqlMatcherActor extends UntypedActor {
             log.info(String.format("Caught exception when processing %s event %s. More info on loglevel 'debug'",
                     eventTypeForLogging, uriForLogging.orElse("[no uri available]")));
             log.debug("caught exception", e);
-            e.printStackTrace();
+            if(log.isDebugEnabled()){
+                e.printStackTrace();
+            }
         }
     }
 
--- a/webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
@@ -26,7 +26,11 @@ export const pricerange = {
   viewerComponent: "won-price-viewer",
   messageEnabled: true,
   parseToRDF: function({ value, identifier, contentUri }) {
-    if (!value || !(value.min || value.max) || !value.currency) {
+    if (
+      value === undefined ||
+      (value.min === undefined && value.max === undefined) ||
+      !value.currency
+    ) {
       return { "s:priceSpecification": undefined };
     }
     return {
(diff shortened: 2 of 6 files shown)


webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
```diff
@@ -149,6 +149,7 @@ public void onReceive(final Object o) throws Exception {
             log.info(String.format("Caught exception when processing %s event %s. More info on loglevel 'debug'",
                     eventTypeForLogging, uriForLogging.orElse("[no uri available]")));
             log.debug("caught exception", e);
+            e.printStackTrace();
```
quasarchimaere: this needs to be removed before merge

webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
```diff
@@ -157,7 +157,7 @@ export const price = {
   ],
   messageEnabled: true,
   parseToRDF: function({ value, identifier, contentUri }) {
-    if (!value || !value.amount || !value.currency) {
```
quasarchimaere: maybe we should handle negative prices here (also?)

webofneeds/won-owner-webapp/src/main/webapp/config/usecases/classifieds.js
```diff
@@ -0,0 +1,243 @@
...
+    // },
+  },
+  goodsDemand: {
+    identifier: "goodsDemand",
+    label: "Look for something",
+    icon: "#ico36_plus",
+    draft: {
+      ...emptyDraft,
+      content: {
+        title: "I'm looking for something!",
```
quasarchimaere: should not be set, default titles could be handled acc. to needtype on viewer/generateHumanReadable

webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
```diff
@@ -44,7 +44,7 @@ export const pricerange = {
         ],
         "s:priceCurrency": value.currency,
         "s:unitCode": value.unitCode,
-        "s:description": "total rent per month in between min/max",
```
quasarchimaere: if we change the desc. here we should change it back to "total rent per month" in the rent search usecase

webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
sigpie: Fixed this and a few other issues with 0 being evaluated as false.

webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
sigpie: The rent range detail uses "Rent in EUR/month" as a label by default and also in the real estate search usecase. I changed this because it makes no sense to have a general price range detail picker that always assumes a specific case.

webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
sigpie: Wrapped it in an if-clause so it actually only prints the trace in the debug log, not the regular log.

webofneeds/won-owner-webapp/src/main/webapp/config/details/price.js
quasarchimaere: ok

webofneeds/won-matcher-sparql/src/main/java/won/matcher/sparql/actor/SparqlMatcherActor.java
quasarchimaere: if clause would be executed if the debug log level is set, but not that it would print it in our debuglog, printStackTrace circumvents the logger entirely

PR #2586, opened 2018-12-13
Refac  loadinghell

### LINKED ISSUE #1571: Automatically reroute when viewing own need url
When visiting your own need url currently a big link to the post view is displayed.
image
Instead, this should be automatically rerouted.

Blocked by #1578

quasarchimaere:
- load/login/logout actions are now easier to read

- Removes the postEnsureLoaded in favour of handling the post in the url in login/logout/load action -> since we need to know if it is actually our own post or not

configRouting changes:
- remove postEnsureLoaded -> did not load a persona in addition to the need
- automatically reroute from visitor view to "main"view if post in uri is own post
- automatically reroute from "main"-view if posturi is still set in the url after logout

- reworked the login/load/action in a way that initialLoadFinished is actually dispatched at the end and not somewhere in between

Fixes #1571 Automatically reroute when viewing own need url


fkleedorfer:
Does not seem to build on integrationtest (see build logs)

quasarchimaere:
build fixed, and deployed on int-test since 2hrs ago

Issue #2599, opened 2018-12-17
Post-Info uses too much space in chat-view, most info unnecessary


quasarchimaere:
![screenshot_20181217-224919](https://user-images.githubusercontent.com/8503486/50117797-387e0b80-024e-11e9-8a9a-77dfaec3fb94.jpg)

the chat-view (which is the default view) uses the space of the screen rather suboptimal (for mobile as well as desktop). there is quite alot of whitespace on the left for example, maybe post-info would be better as a fold out (which is expanded by default for certain connection states) or something else (ux-perts @sigpie @peacememories @m0ru probably know how to tackle this).

also (as seen in the screenshot) most of the information in the post-info header is rather unnecessary (nobody knows what flags are, types are cryptic (incl. the omnipresent won:Need), the share text is also very in your face as is the whole share dialog itself (imho sharing is already linked as a single well understood share-button which could in return lead to a more specific sharing option-view but what is displayed now is using too much of the screen...)

might be related to #2597

PR #2637, opened 2019-01-11
Feat less cluttered post content

sigpie:
Contributes to #2599 

Hides Facets, Flags, Matching Context and some Type details by default. All information is shown if `showRdf` is enabled, as most of this information is only needed for debugging purposes or power users.


quasarchimaere:
i would not merge this, some facets (eg groupfacet) and some flags (matchingflags) as well as the matchingContext are very valuable information... not only to the developer but also to the user, i agree that we might want to reduce the clutter by hiding some of the content but we should not hide it in general, we might also be able to find better labels for the items

sigpie:
As far as I can tell, there's two flags and one facet that might be of interest for users:

* No Hints for counterpart - maybe show this as "Invisible"
* No Hints for me - maybe show this as "No Matches"
* Group Facet - maybe show this as "Group Chat enabled"

Are there any other flags that should be always visible?

sigpie:
Force-pushed after rebasing on current master.

sigpie:
Changed behaviour to include matching context, group chat facet and "No hint for..." flags.

PR #2660, opened 2019-01-21
Fixes post-visitor view and viewNeedUri-modalbox view for deletedNeeds

### LINKED ISSUE #2630: Show information text on deleted posts
If a deleted need is opened (e.g. via a share-link that was sent before the need was deleted), an empty need is shown with no indication on why there is no visible content. In this case, some information message should be displayed, such as "Remote Need Loading failed".

### LINKED ISSUE #2631: Disable interaction if remote need could not be loaded
If the remote need could not be loaded, e.g. because a suggested need got deleted, the controls should be disabled. For the need shown below, this applies to the "Ask to Chat" button and the share-menu. 

deleted_need_available

quasarchimaere:
Show loading indicator while loading, and failedToLoad info if the post was not able to be fetched

fixes #2630 

How to Test:
- open a visitorlink of a deletedNeed -> you should see that the need has been deleted
- open a visitorlink of a need that takes long to load (e.g. a need with images) -> you should see a loading indicator until the need has been loaded
- open a deleted post by adding &viewNeedUri=[deletedNeedUri] into the connections url -> this way you will mock clicking on a suggestedPost and see that the need has been deleted in the overlay
- look at a hint of a deletedNeed (e.g. if a whatsNew match could not be loaded) -> see that the click interaction to open the right side (view connection) was removed -> fixes #2631

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-selection-item.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/connection-selection-item.js
@@ -22,6 +22,7 @@ function genComponentConf() {
       <won-connection-header
         connection-uri="self.connectionUri"
         timestamp="self.lastUpdateTimestamp"
+        ng-disabled="self.remoteNeedFailedToLoad"
         ng-click="self.setOpen()"
         class="clickable">
       </won-connection-header>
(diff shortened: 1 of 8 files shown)


fkleedorfer:
Change request regarding wording: use 'Post' instead of 'Need'. 
Instances: 
* Failed To Load - Need might have been deleted 
* Remote Need Loading Failed
* Need might have been deleted
Plus: please remove 'You might want to close this connection' from connection list view

PR #2669, opened 2019-01-22
Fix slideinheight #2649

### LINKED ISSUE #2649: Slide-in is not counted for determining content height
Having a slide-in visible means that the footer is always out of sight, even if the content would fit on the screen. This is due to the slide-in height not being counted when determining the needed content height to avoid scrolling. 

It's annoying to have to scroll to see the chat-input and footer while a slide-in is visible. Possible fixes would be a close button for the slide-in or changing the content height calculation to allow for the slide-in.

quasarchimaere:
Fixes #2649 

- Slide ins are now collapsible as a whole (indicator is shown in the topnav if that is the case)
- Also adds loadingIndicator in the topnav, to show if ANYTHING is currently being processed


sigpie:
Nitpick: `login-form,js` should use `isProcessingResendVerificationEmail` as a function name instead of `processingResendVerificationEmail` for more consistency. 

Further nitpick: `termsOfServiceAccepted` has been only partially refactored to `isTermsOfServiceAccepted` and seems not to have an utility method for getting this state.

sigpie:
Slider content can overlap:

overlap

sigpie:
Getting a lost connection slide-in hides all other slide-ins. Other slide-ins only become visible again after a reconnect.

quasarchimaere:
> Getting a lost connection slide-in hides all other slide-ins. Other slide-ins only become visible again after a reconnect.

this is intentional (mimics the old behaviour)

quasarchimaere:
> Nitpick: `login-form,js` should use `isProcessingResendVerificationEmail` as a function name instead of `processingResendVerificationEmail` for more consistency.
> 
> Further nitpick: `termsOfServiceAccepted` has been only partially refactored to `isTermsOfServiceAccepted` and seems not to have an utility method for getting this state.

i agree, i didnt refactor all of the occurences for which we have helper methods yet, because i wanted to not bloat this PR too much

sigpie:
Currently weird UX: hiding a slide-in, scrolling down and showing it again seems to remove the slide-in as it's added above the currently visible part. This will be fixed automatically if slide-ins are attached to the title bar and are shown above other content in a future iteration. Alternatively, the warning sign in the header could stay visible after expanding the slide-in.

quasarchimaere:
resizing should now only happen on really small devices with the disclaimer visible, idk how to fix this right now but i assume that we can fix it later

sigpie:
The "Link copied to clipboard. X" message from the anonymous account slide-in stays visible after folding and expanding the slide-in - as closing this message and expanding again results in the message being reset, manually folding and expanding should do so, too.

PR #2673, opened 2019-01-22
Make parsing errors of WonMessages more verbose

fkleedorfer:
This commit adds the graph URIs to the error message.
The graph URIs are the names of the graphs found in the rdf document
that we attempted to parse as a message and failed.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
@@ -1768,17 +1768,27 @@ WonMessage.prototype = {
     }
 
     if (innermostEnvelopes.length == 0) {
-      this.parseErrors.push("no innermost envelope found");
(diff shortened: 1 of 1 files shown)


webofneeds/won-owner-webapp/src/main/webapp/app/service/won.js
```diff
@@ -1768,17 +1768,27 @@ WonMessage.prototype = {
     }
 
     if (innermostEnvelopes.length == 0) {
-      this.parseErrors.push("no innermost envelope found");
```
quasarchimaere: instead of doing any of this please just print the graphUris as part of the console.warn within this file line#:918 that way we do not have to dabble with the stringified version of the graphUris `wonMessage.graphUris`

quasarchimaere:
close this pr in favor of verbose logging within the "constructor" of the wonMessage

PR #2675, opened 2019-01-23
Feat  verbose wonlogging

quasarchimaere:
Adds reference to access parseMessage(wonMessage) method from the console

parseMessage4dbg -> 3 params, `parseMessage4dbg(wonMessageObject, boolAlreadyProcessed, boolForwardMessage)`
- the 2 bool parameters are more or less irrelevant, `boolAlreadyProcessed` sets the read flag to true by default `boolForwardMessage` returns a message that looks like a forwardedMessages (certain attributes are different but all in all it doesnt really matter to check parsing)


sigpie:
Still losing some group chat messages. Also got an unexpected double forward:

double-forward

Matches are suggested to the group chat need, but requesting someone to join from the group chat need seems not to work properly. The other side does not get a request and if the other side sends a request, the group chat admin interface needs a reload to correctly display the request. Rejected requests also stay until a reload.

PR #2703, opened 2019-01-29
Feat uc cycling

fkleedorfer:
Adds two use case groups: cycling and lunch
both are identical, except for a few strings - so there is refactoring potential.

The idea of both use case groups is that you post an 'interest' only once, to signal that you would be up for doing it, and when you actually want to do it, you post a 'plan', which matches all the interests (if their filter criteria are met)

Minor improvements were done alongside this
* replace usages of won:location with s:location
* a unit test was extended


fkleedorfer:
Food for thought: 
* Joining groups is still awkward, especially for the group owner. 
* having to accept group members is tedious
    * if we had buddy relationships, we could provide auto-open only for buddies. Similar for other relationships

quasarchimaere:
merging this pull request would result in any pre-pull created need/message with location detail to become invisible since we do not check won:location any longer, i assume that this would also make our existing job needs or at least the existing real estate demo needs unusable... if thats ok thats ok but maybe we should allow a bit of resilience here as seen in the title or description parseFromRdf

PR #2714, opened 2019-01-31
Start from Template

### LINKED ISSUE #2702: me too functionality
On the need detail view and visitor view, allow users to take the need data and repost it as their own, after optionally modifying it.

quasarchimaere:
fixes #2702 

Allows the user to create a need based on an existing one:

Caveats:
- does not retrieve the correct useCase based on the need yet (we need to set a type in each usecase to find the correct one)
- only parses details set in customUseCase
- shows all details in customUseCase

... works though


fkleedorfer:
Bug: not working from visitor view - details are not filled in when clicking 'Post this too!'

PR #2727, opened 2019-02-04
Removing needs from `held` when deleting.

### LINKED ISSUE #2652: Deleted Need still shows up in derivedData of a persona (holds needUri)

peacememories:
Fixes #2652


quasarchimaere:
the pr does not fix the issue at hand, see comment above for details

peacememories:
Propagating all the necessary changes to the client might prove a challenge. If, for example, a persona is loaded that resides on a different node, we won't know about the deletion of an associated need with our current system.
We'd need to create an update channel to any node we have data from.

PR #2739, opened 2019-02-07
Feat  start from usecase #2715

### LINKED ISSUE #2715: Start From Template should use correct useCase
We need to add a specific set of types within the content of each usecase so we can extract the proper usecase to use within create-post.js (basically we need to find the useCase by "fromNeed")

~~If we can't find the proper useCase then we should extract all the set details (seeks and content) from the "fromNeed" and display their respective pickers instead -> this should not be happen alot though, be aware that the details to choose From might hide within a specific useCase (not all the available details are within the customUseCase)~~
will be done within a different issue, lower priority since it should not happen after this is fixed anyway
blocked by #2714

### LINKED ISSUE #2720: Post this too! does not generate embedded query
When createing a need via 'post this too', the hasQuery property isn't set

requires #2715 (and if that is implemented this issue should have been resolved itself)

quasarchimaere:
Fixes #2715 -> find a matching UseCase for a given need and set the create view to the appropriate input fields, if multiple matches are found the customUseCase will be used instead.
Fixes #2720 -> if a useCase is found we also generate the defined query now

Adds a somewhat "sane" type to all usecases so we can easily find a matching usecase for a given need

### DIFF
--- a/webofneeds/won-owner-webapp/README.md
+++ b/webofneeds/won-owner-webapp/README.md
@@ -200,6 +200,11 @@ $ngRedux.getState();
        unread: true|false, //whether or not this need has new information that has not been read yet
        uri: string, //unique identifier of this need
        humanReadable: string, //a human Readable String that parses the content from is or seeks and searchString and makes a title out of it based on a certain logic
+       matchedUseCase: { //saves a matchedUseCase within the need so we dont have to parse it multiple times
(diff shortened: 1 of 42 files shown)


quasarchimaere:
closing for now, found a bug

quasarchimaere:
bug is fixed reopening the pr

PR #2757, opened 2019-02-18
Feat  enhance postcontent

quasarchimaere:
First implementation of cleaning up/enhancing the post-content view
- Implements Tabs to reduce visual clutter for the default view
- Directly go to suggested Tab if "suggestions" of a need is clicked (within the connection overview)
- Directly go to detail Tab if Details-Button is clicked (within the connection overview)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
@@ -157,6 +157,7 @@ const actionHierarchy = {
     storeUriFailed: INJ_DEFAULT,
 
     toggleGeneralInfo: INJ_DEFAULT,
+    selectTab: INJ_DEFAULT,
   },
   personas: {
     create: createPersona,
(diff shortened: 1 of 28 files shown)


sigpie:
This makes suggestions initially invisible - At first glance, there seems to be no interaction at all on posts, even though there are suggested matches. The only indication is a black arrow for showing details instead of a grey one.

sigpie:
Marking suggestions as read is annoying - need to view details of each and move pointer to close the detail view again. Maybe they should be marked as seen a short time after opening the tab.

sigpie:
As this shows post creation time instead of matching time, it's more visible that there is no useful sorting of suggested matches, as seen below. New posts are mixed with week- or month-old posts, even on What's New. This might be a general issue, not an issue with this PR though.

sorting

quasarchimaere:
> This makes suggestions initially invisible - At first glance, there seems to be no interaction at all on posts, even though there are suggested matches. The only indication is a black arrow for showing details instead of a grey one.

is this really the case? i am asking because if so it is a won-unread indicator bug, because there should be an indication bg-color that indicates that there is something unread within this element (in this case unread suggestions)

sigpie:
> is this really the case? i am asking because if so it is a won-unread indicator bug, because there should be an indication bg-color that indicates that there is something unread within this element (in this case unread suggestions)

bg-color changes as expected, but is not a good unread indicator. Additionally, there's a bug for unread suggestions: open need -> remove suggestions without opening them -> eternal unread indication that only disappears after a reload.

PR #2760, opened 2019-02-19
Conf file cleanup

sigpie:
Removed some conf files and tests - since most of the deleted files have not been accessed in years, everything should still work as expected. This might require changes to your local server config, as all logging configurations except `logback.xml` were deleted.


sigpie:
closing this for now, further testing needed

PR #2800, opened 2019-03-12
remove saving closed connuris in localstorage and use the correct state from the backend (Feat lds uris by connstate)

### LINKED ISSUE #2778: Reopened connections break state
When a connection is reopened, say, by attaching a persona to a need that was previously attached to it, the derived data belonging to that connection gets loaded correctly, but since the connection is blacklisted from loading in local storage, this results in inconsistent state.

Removing this blacklist would have a strong negative impact on performance.

quasarchimaere:
fixes #2778 

Frontend changes:
- removes the closedConnUri check from localStorage (closedConnUris wont be stored in localStorage at all)
- implements a new method that retrieves not only the connectionUri but also the state
- only add active Connections based on their real state from the server and not from localStorage

Backend changes:
- removed unused methods
- connectionContainer updates -> do not only give a list of connUris but also their state and facets and additional info

### DIFF
--- a/webofneeds/won-core/src/main/java/won/protocol/repository/ConnectionRepository.java
+++ b/webofneeds/won-core/src/main/java/won/protocol/repository/ConnectionRepository.java
@@ -162,6 +162,14 @@ public interface ConnectionRepository extends WonRepository<Connection>
     @Param("referenceDate") Date referenceDate,
     Pageable pageable);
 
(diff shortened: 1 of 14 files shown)


quasarchimaere:
merging possible, but not recommended before the demo tomorrow (on march 13th)

quasarchimaere:
removed the do not merge label, demo is complete afaik

PR #2815, opened 2019-03-13
fix textfield focus color to be browser independent (fixes #1813) 

### LINKED ISSUE #1813: Different text field focus colours
When focusing a text field on desktop it's blue: focus_blue

But when focusing the same text field with the user agent set to a mobile device, it's orange: 
focus_orange

quasarchimaere:
add :focus outline color to be won-secondary-color by default (should overwrite any browser specific focus behaviour)

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/style/won.scss
+++ b/webofneeds/won-owner-webapp/src/main/webapp/style/won.scss
@@ -114,6 +114,10 @@ input[type="email"] {
   border-radius: 0.19rem;
   border: $thinGrayBorder;
   font-size: 16px;
+
+  &:focus {
+    outline-color: $won-secondary-color;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


sigpie:
Does not work on all textfields. 

* chat message input has no highlighting colour
* description detail and time/date detail still use the system colour

quasarchimaere:
> Does not work on all textfields.
> 
> * chat message input has no highlighting colour
> * description detail and time/date detail still use the system colour

fixes the second part of the comment -> overwrites system color for date/time/url/number inputs and textareas and select fields.

WON'T FIX / DOES NOT IMPLEMENT: Highlight color for the chat input field -> never had one and does not make sense to show one since there is only one input field visible and also the default "state" is selected if you will, and that would result in an outline being shown all the time, which would be weird

PR #2826, opened 2019-03-14
Feat change password backend

### LINKED ISSUE #2563: Change password option in settings
Would be a good idea for better usability.

Related to #264 
Related to #363 

child of #2806

ask @fkleedorfer about the todo from #2826
resetPassword in RestUserController now does not check the verificationToken. This should be added (requiring an email roundtrip before reset is allowed)

### LINKED ISSUE #2812: Change password backend
child of #2806

fkleedorfer:
Fixes #2812 
# Backend
Password change back-end works, keystore password is also changed
An Email notification is sent to the user

# Frontend
use new `changePassword` function in `won-utils.js`
## Still TODO
* [ ] requires #2563 
* [ ] resetPassword in RestUserController now does not check the verificationToken. This should be added (requiring an email roundtrip before reset is allowed)


quasarchimaere:
whoever implements #2563 shouldnt forget to include the new response statuscodes, introduced here, into won.js

Issue #2833, opened 2019-03-19
Offer the option to login within the termsDialog


quasarchimaere:
We should offer a login directly within the showTermsDialog so a user can login from there and execute the correct action after the login was processed.

related to #2537
requires #2834

quasarchimaere:
moved to our internal idea board

Issue #2527, opened 2018-11-21
Fix Scrolling issue in new about page


quasarchimaere:
The Scrollposition of the new aboutpage is remembered somehow and thus resulting in a very uncomfortable UX

requires #2523

PR #2834, opened 2019-03-19
Feat  agree tos private #2537

### LINKED ISSUE #2537: Agree to ToS for privateid users
Even though we currently do not save the TermsOfService Agreement flag on the server we still have the flag within our signup gui.

however since it is possible to create needs without an account (privateIdUsers) we might have to include the acknowledgement of the terms of service whenever a (not yet) user tries to create a need.

since we have a modal dialog we could facilitate it for this instance as well:
show the modaldialog before any need creation that doesnt have a user yet (and would create the user during need creation)

ModalDialog text should be something like "By publishing/creating this post you agree to the termsofservice(link)"
with two buttons "proceed" "cancel/nevermind" (see the close need action in the post-context-dropdown) to find out how modal dialogs work.

we need to add this for every createpersona, createpost, createsearch, createwhatsaround, createwhatsnew actions

this will also be necessary for tos-updates that reset the serverside(pending implementation) flag for the users.

the check if the modaldialog is NOT necessary will be stored as a flag within the user in the state, probably something like user->agreedToTerms, so whenever this flag is set to false or is not present we would need to show the modaldialog.

tbd: if logged in users should automatically opt in if they proceed with the need creation or if it should be blocked until the user has specifically agreed to the terms in some other way.

quasarchimaere:
Adds a modal-dialog before executing the following actions in the following places (this is also how/where to test) when not logged in:
- publish in create-post
- publish in create-search
- createWhatsNew in usecase-picker
- createWhatsAround in usecase-picker
- createWhatsNew in about page
- createWhatsAround in about page
- Send (AdHoc)-Request from visitor view.

All these actions display a "Attention- if you proceed you accept the tos"-dialog now,
no just closes the dialog
yes continues with corresponding action

fixes #2537 

we could implement a login dialog as another option within this new dialog -> this is a different issue though (see #2833)

If there are issues with the wording of the dialog please change directly within this pr:
see `modal-dialog.js` component

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/account-utils.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/account-utils.js
@@ -89,3 +89,12 @@ export function getEmailVerificationError(accountState) {
 export function isDisclaimerAccepted(accountState) {
   return !!get(accountState, "acceptedDisclaimer");
 }
+
+/**
+ * Return true if the TermsOfService are already accepted
+ * @param accountState
+ * @returns {boolean}
+ */
+export function isTermsOfServiceAccepted(accountState) {
+  return !!get(accountState, "acceptedTermsOfService");
+}
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/account-utils.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/account-utils.js
@@ -10,7 +10,7 @@ import { get } from "./utils.js";
  * @returns {*|boolean}
  */
 export function isLoggedIn(accountState) {
-  return get(accountState, "loggedIn");
+  return !!get(accountState, "loggedIn");
 }
 
 /**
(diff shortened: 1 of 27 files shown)


fkleedorfer:
Clicking on the terms of service link does not jump to the right place in that page. The URL looks broken:
`https://localhost:8443/owner/#!/about#aboutTermsOfService`

quasarchimaere:
url should open in new tab, will have a look at it

fkleedorfer:
It does - but it does not scroll to an anchor, as seems to be intended

fkleedorfer:
also: typo in the message: anoymous -> anonymous

quasarchimaere:
> It does - but it does not scroll to an anchor, as seems to be intended

see issue #2527

quasarchimaere:
> also: typo in the message: anoymous -> anonymous

see last paragraph of pr description

quasarchimaere:
> Clicking on the terms of service link does not jump to the right place in that page. The URL looks broken:
> `https://localhost:8443/owner/#!/about#aboutTermsOfService`

but you are right the url is broken, look at the one in the footer (`https://www.matchat.org/owner/#!/about?aboutSection=aboutTermsOfService`) this one is correct, i assume that since i copied the url generation from the signup page we might want to edit the url there as well, and also check the url wothin the terms of service slide-in (appears if logged in with a user that hasnt accepted the terms of service yet)

fkleedorfer:
> see last paragraph of pr description

I read that - but I wanted to avoid an edit conflict.

quasarchimaere:
updated text according to discussion with @fkleedorfer , fixed terms of service link on the signup page and within the terms dialog

PR #2828, opened 2019-03-18
Feat store connection state and (own) facettype in redux-state onload #2823

quasarchimaere:
blocked by #2800 

this pr stores the connection state and the (own) facet(type) of each connection directly after retrieving the connection container, this makes it easier to reload or prioritize loading by connectionstate and/or facet

fixes #2823

PR #2841, opened 2019-03-21
Feat store connection state and (own) facettype in redux-state onload #2823  (rebased)

### LINKED ISSUE #2823: Save facets/connectionstate to reduxState when storing the uris
requires #2800

Now that we retrieve the connectionState as well as the facets directly within the connectionContainer, we can add the additional data directly into the rdf store and thus can enhance the performance of (initial)load or login by prioritizing and/or omitting retrieval of connections based on the additional information

This also allows us to remove the somewhat odd ui quirks where connections inLoading show up as skeletons but disappear after the connection was loaded

### LINKED ISSUE #2840: Conversation messages not shown after reload
But on PR #2828

Note: Bug observed when PR is rebased onto current master (addition from @quasarchimaere -> branch was not tested before rebase, so the issue does not necessarily have anything to do with the rebase)

Steps to reproduce:
1. have a conversation, e.g. with the debug bot
2. while viewing the conversation, reload
--> messages are not reloaded, the loading spinner stays forever

quasarchimaere:
~~blocked by #2840~~

related to #2828 (same changes just rebased with current master)

this pr stores the connection state and the (own) facet(type) of each connection directly after retrieving the connection container, this makes it easier to reload or prioritize loading by connectionstate and/or facet

fixes #2823
fixes #2840

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
@@ -131,6 +131,7 @@ const actionHierarchy = {
     updatePetriNetData: INJ_DEFAULT,
 
     storeActiveUrisInLoading: INJ_DEFAULT,
+    storeUrisToLoad: INJ_DEFAULT,
     storeActive: INJ_DEFAULT,
 
     storeUriFailed: INJ_DEFAULT,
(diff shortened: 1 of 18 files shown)


quasarchimaere:
was blocked by #2840 -> fixes/changes pushed directly into this branch

PR #2850, opened 2019-03-27
Add java formatting to build

fkleedorfer:
Finally builds the same under windows and linux.

The build now validates the formatting against the eclipse formatting rules in 
```
won-buildtools/src/main/resources/formatter.xml
```
Changes to those rules only become effective after doing a `mvn install` in `won-buildtools` (which happens during the normal maven build of the overall project)

Troubles with the build failing for jenkins are now resolved.


sigpie:
I was unable to build before running `mvn formatter:format` on the whole project, and commiting the "changes" - no files were actually changed, but after this, `mvn install` succeeded.

quasarchimaere:
i was also unable to get this to work:
- tried to check out branch without any changes -> mvn install fails because it is not formatted (acc. to the mvn install output)
- tried to do mvn install of won-buildtools before, the full mvn install -> still fails
- tried to execute formatter on the whole project, basically every file is affected, but...
  - changed files show no changes at all
  - local build runs now,
  - trying to commit a file result in the following error: `Commit failed with error:
warning: LF will be replaced by CRLF in webofneeds/won-node/src/main/java/won/node/facet/impl/AbstractBAFacet.java.
The file will have its original line endings in your working directory.`

i guess this is due to the system independent new line character check, which is ok with me but i feel like the correct formatting rule of these new lines seems to be forgotten after checkout... also i think that the settings.xml of the formatter only sets non default values (e.g. can't find the max line width within the config).... not sure if all ide's can handle using eclipse default formatting rules if they arent specified... to be frank i dont like this, or the handling that follows after (can't forget to call formatter before build, can't forget to build on the local machine before commits to ensure that the build will def. not break on jenkins... not sure if any formatting that was applied locally will not result in breaking the build anyway... and so on.... this formatting should enhance the DX and not make it more complicated...

quasarchimaere:
Added the do not merge flag, since this pr does not build (tested by @sigpie and @quasarchimaere)

quasarchimaere:
> i was also unable to get this to work:
> 
> * tried to check out branch without any changes -> mvn install fails because it is not formatted (acc. to the mvn install output)
> * tried to do mvn install of won-buildtools before, the full mvn install -> still fails
> * tried to execute formatter on the whole project, basically every file is affected, but...
>   
>   * changed files show no changes at all
>   * local build runs now,
>   * trying to commit a file result in the following error: `Commit failed with error: warning: LF will be replaced by CRLF in webofneeds/won-node/src/main/java/won/node/facet/impl/AbstractBAFacet.java. The file will have its original line endings in your working directory.`
> 
> i guess this is due to the system independent new line character check, which is ok with me but i feel like the correct formatting rule of these new lines seems to be forgotten after checkout... also i think that the settings.xml of the formatter only sets non default values (e.g. can't find the max line width within the config).... not sure if all ide's can handle using eclipse default formatting rules if they arent specified... to be frank i dont like this, or the handling that follows after (can't forget to call formatter before build, can't forget to build on the local machine before commits to ensure that the build will def. not break on jenkins... not sure if any formatting that was applied locally will not result in breaking the build anyway... and so on.... this formatting should enhance the DX and not make it more complicated...

the commit fails because the commit is "blocked" by the precommit hook (that anybody gets when installing the owner-webapp with `npm install` (afaik) and imho thats a good thing...

fkleedorfer:
> the commit fails because the commit is "blocked" by the precommit hook (that anybody gets when 
> installing the owner-webapp with npm install (afaik) and imho thats a good thing.
How is that possible?

Also, the LF/CRLF problem seems to be git-installation specific. We could use repo specific settings: https://help.github.com/en/articles/dealing-with-line-endings

@quasarchimaere  could you try your commit again and see if the problem persists?

Also, how about we don't run the validate goal in the normal build, but we still have the formatting rules and the maven goals so we can format/validate if we like - that should be good enough.

quasarchimaere:
i added the .gitattributes file to the base of the project, this ensures that any new checkout will use LF as the eol characters, this should ensure that mvn install (or to be specific mvn formatter:validate) will work on non unix machines.

See https://help.github.com/en/articles/dealing-with-line-endings for help, if one still experiences validation problems after a checkout, you can reset the git index by deleting the `.git/index` file within your repository and execute `git reset` after to rewrite any files that where still using the wrong eol character(s)

PR #2857, opened 2019-04-01
Added login check to settings route.

### LINKED ISSUE #2829: Account Export seems to be possible while being logged out
How to reproduce:
1) view any need with the visitor view
2) go to "personas" from the "Create Personas" button within the publish as dropdown
3) go to Account -> see that export is visible and that you can actually execute it as well -> "Export Started. You will get an e-mail soon" -> even though there was nothing to export, -> WS Call to "exportAccount" returns a 302 (found) and no error at all

peacememories:
Fixes #2829

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/configRouting.js
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/configRouting.js
@@ -164,6 +164,13 @@ export function accessControl({
       return;
     }
 
+    case "settings":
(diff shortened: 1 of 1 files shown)


fkleedorfer:
Having trouble reproducing #2829... skipping this PR for now

quasarchimaere:
not sure if the reprosteps described in #2829 make sense... from what i see in the code though, one could reproduce this by having many needs and then try to access the settings page while initialload is still running, that should enable the change to the settings page without being logged in (on master) this pr prevents this from happening

Issue #2795, opened 2019-03-11
Elm components display persona icon different to the rest of the application


quasarchimaere:
Persona icons within all elm component display the persona (ident)icon as squares, whereas the rest of the application displays persona-identicons as circles (to distinguish between needs and personas better)

![personapicker](https://user-images.githubusercontent.com/8503486/54110212-175a3080-43e1-11e9-8f78-cd7fdcbd7bb1.PNG)

quasarchimaere:
partially fixed by #2882 -> the settings view does not have circular persona icons though so this is still to be done

Issue #2792, opened 2019-03-11
Persona Selection will not work for Users with many Personas


quasarchimaere:
Persona selections are represented within dropdown fields, which won't work well if a user has many personas

![personapicker](https://user-images.githubusercontent.com/8503486/54110067-b16da900-43e0-11e9-9aaa-224dedfc2790.PNG)

the amount of visible personas is also dependant on the height of the parent component (if a parent need info does not contain alot of info even less personas are shown)

quasarchimaere:
better due to scrollable list, but still happens in creating (from) usecases with few detail pickers -> ReproSteps:
- logged in with a user with many personas (~10)
- access Create "Contact Payment Bot"-Usecase (located in "More")
- open the persona selection
- see that list scroll out of the view and the top part will remain inaccessible

PR #2882, opened 2019-04-17
Refactored Elm components

### LINKED ISSUE #2791: Add Persona "interaction" different than other persona buttons
the add persona "dropdown" within a need adds the persona as soon as you select one (after the click of add persona).

This is a different interaction sequence compared to the other add-persona (or post as persona) interactions and possibly confuses users.

### LINKED ISSUE #2793: "Add Persona"-Picker in Need has different styling than other persona pickers
See comparison in line height between "add persona" to need and "post as"- view

personapicker
personapicker_create

requires #2795

### LINKED ISSUE #2794: Persona List(s) unsorted (in persona overview, and dropdowns)
The persona list is not sorted by name, only by creation date, needs to be sorted by name

### LINKED ISSUE #2798: "Add Persona" does not show a clear "processing" or "in progress" feedback
Adding a persona takes a couple of seconds from select click to the persona appearing as the holder of the need

instead of assuming or proactively adding the persona to the need, we wait until the corresponding ws-calls have arrived and are processed correctly, even though this behaviour is not wrong it is still somewhat unclear to the user that anything happens during that process, maybe we need to add an indication that a persona is currently being added/or removed to/from the need

### LINKED ISSUE #2804: Removing a persona within the persona tab results in style errors
Removing a persona from within the persona tab results in not displaying the personatab any longer, and removes the persona that was once visible, however the tab does not switch to a sane default (e.g. details tab) on click or after the removal has been processed, resulting in the addpersona button being visible again, showing no tab visible at all, and adding a scroll bar to the parent element of the add persona button again

### LINKED ISSUE #2807: "Add Persona"-button issues when not having any personas
- "Add Persona" button is shown even when user does not have a persona: Two options, either remove the disabled interaction and show the "Create persona"-element (similar to the publish button), or remove the button altogether if no persona is available to be added anyway
- the button is also visibly inactive but still has the pointer cursor on hover indicating a clickable element

peacememories:
Fixes #2793
Fixes #2807
Fixes #2804
Fixes #2791
Fixes Partially #2792
Fixes #2798
Fixes #2794
Fixes Partially #2795


quasarchimaere:
whats the partial fix in #2795

PR #2910, opened 2019-04-30
Feat  enhance whats around

quasarchimaere:
Implements a search-input field in the whatsAround view, that way you can search around any location and retrieve the atoms that are around the location.
This enables users who did not grant geolocation access to still see whats around certain places


sigpie:
What's around default zoom level is too high if no matches were found. By default, at least the search radius should be visible, even if there are no results. 

The map periodically resets its zoomlevel and position, even while it is moved by the user. This should not happen. Instead of an automatic recentering, a "reset view" button could be added.

PR #2926, opened 2019-05-13
Fix description of won:textMessage

fkleedorfer:
Fixed the description of won:textMessage in the ontology. Also found out why we cannot move it to msg.


quasarchimaere:
ok i understand that hasTextMessage is not allowed in msg due to message validation, merging this pr (@fkleedorfer should i add an issue to replace won:hasTextMessage with some preexisting ontology of schema.org?)

PR #2930, opened 2019-05-15
Switched js build system and removed html dependencies.

peacememories:


quasarchimaere:
Do not merge:
- `overview.jsx` differs too much from the `overview.js` that was deleted (look at the file in the master branch) -> template (from `overview.html`) and Controller (from `overview.js`) need to be properly integrated in the new overview.jsx file
- Integration-Test doesn't show the WelcomeText (above the usecase picker, for whatever reason... look at int test owner to see what i mean) seems to me that the "welcomeTemplate" is not retrieved properly -> maybe the `configInit`-function in the `config-actions.js` is not executed properly or tries to access directories that are not in the bundle or idk (what i mean is that it seems to me that the static resources/skins are not retrieved properly)
- Build does not work on windows environments `package.json`  `"scripts": {
    "build": "WON_OWNER_BASE_URL=/owner parcel build index.html --public-url=/owner -d generated",
    "serve": "parcel index.html",
    "postinstall": "npm rebuild && patch-package",
    "precommit": "lint-staged",
    "watch": "WON_OWNER_BASE_URL=/owner parcel watch index.html --public-url=/owner -d generated",
    "clean": "rimraf ./generated ./.cache"
  },` the `build` and `watch` goal do not work since `WON_OWNER_BASE_URL` is interpreted as a command and not a var (`npm run watch` and `npm run build` need to be env independet as it was before)

This PR needs to be checked on a windows machine locally after fixes have been made, and only if it builds on the windows environment as well as the unix env (e.g. integration-test) we can merge.

quasarchimaere:
Also, at least on my machine the build did not run because it said some random .elm file is currently accessed by another process, (i changed the build script so it works on my windows machine locally)
i dont know if this is caused by our antivirus in the domain or bc there are issues in the parcel build (webpack did not have this issue) still, even if it is due to our antivirus and not parcel itself we need to find a solution that works for our dev-team as we will def. not get the opportunity to fully switch to unix (at least not every member of the team)

PR #2943, opened 2019-05-21
Refac  socket handling #2940

### LINKED ISSUE #2940: Refactor Socket Handling
We currently store connections and their respective sockets by fetching the correct socket based on the socketUri of the connection, this leads to a state where we have to have both atoms stored in the state before we can store the connection (since the socketUri <-> socket is only stored within the atom). This used to be a non-issue in the past as we were able to fetch both atoms before adding the connection, since the change to socketHints this is no longer possible (targetAtom can only be fetched after the connection has been fetched)

We could fix this in 2 ways:
1. leave the state as is, but add another roundtrip to store the correct sockets in the connection after both atoms have been fetched
2. store the known socketUris alone and retrieve the sockets based on the socket map keys in the (stored) atoms

I will implement approach 2 as this reduces the round trips and is cleaner than the other approach (less data duplication)

quasarchimaere:
fixes #2940 

Since we always know which two sockets are connected within a connection now, we can eliminate the assumed remoteSocket of a connection.

This Pull Request refactors the redux connection state so that only the socketUris are stored, and not the realSockets (which are based on existing/present atoms in the state) -> see description of #2940

How to Test (probably best to deploy on integration-test):
- create a "Search For a Job"-Atom
- see that suggested connections show up like they used to


quasarchimaere:
do not merge flag added, groupChat participants handling seems to be off

quasarchimaere:
Removed do not match flag, connections involving groupSockets should work as intended

quasarchimaere:
Merging this PR on my own, found out that the current master does not support groupChats due to the refactoring in prior pull requests

PR #2955, opened 2019-06-11
Feat fetch pokemon from service

sigpie:
Known issues: 
* a few pictures ~~mostly~~ don't work, this is due to the links returned by the listservice
* ~~pokemon viewer currently can't display all pokemon, this is due to the viewer using the fallback list to retrieve information instead of its own fetch call - to fix this, we need to either update the fallback list (necessary anyway) or refactor pokemon viewing to also use the listservice~~ fixed
* ~~pokemon viewer uses placeholder icons until fixed pokemon picture links are available~~ fixed

ADDITION:
* ~~the fetching of pokemon raids has revealed that "level" is not necessarily an attribute that is present -> if a pokemon has already hatched(e.g. the raid Started) there might not be a level present, so please let us change the picker and viewer as well as the detail validation according to this requirement~~ fixed


sigpie:
Added Do Not Merge until new pokemon pictures have been tested and added

PR #2963, opened 2019-06-15
RDF-Structures for the reputation process

mszkb:
Implementing RDF-Structures as Junit tests for the reputation process.


mszkb:
I need to change the type of the certificate property inside the reputation token.

quasarchimaere:
ok

quasarchimaere:
close without merge, utils and vocabulary will be implemented in the project (https://github.com/mszkb/won_reputation_privacy) directly (for now)

PR #2976, opened 2019-07-02
Feat  spoco updates

quasarchimaere:
Adds abstract helper class to simplify AtomDeletion Actions in bots

Adds missing Resource/Property constants in SCHEMA.java (e.g. PlanAction)

Avoids checking the signature for successResponses of DeleteMessages (signature can't be checked since the signature has been deleted)


quasarchimaere:
tested on integration-test and with spoco raid-bot

PR #2982, opened 2019-07-03
Fixes #2967

### LINKED ISSUE #2967: Signature failure when sending TTL messages
Messages sent in RDF mode, e.g. `<this:eventuri> con:text "hello world!".` cause an exception and as a result cannot be sent. This means that the owner application currently cannot be used to send custom TTL messages.

sigpie:
Fixes TTL parsing in RDF mode.

fixes #2967


fkleedorfer:
Tested with ttl messages as well as location, start date and price details.

PR #3000, opened 2019-07-10
Fix  useratom insert error

### LINKED ISSUE #2988: Posting 'Plan a ride' very unreliable
Steps to reproduce: 
* have Persona
* have cycling interest
* create a ride (->'plan a ride') using the persona.

Trying this, (in ff and chrome), a screen like the following is not unusual:
grafik

It takes quite a long time until the server-side action for accepting the holder connection is sent (I understand this may or may not have to do with the owner webapp). Even after that, the view remains broken and a reload is necessary. 
The expected result (a new ride within the persona's inventory) is not always achieved. Sometimes, no new atoms are created (or at least shown in the GUI)

fkleedorfer:
Fixes #2988 

WRT #2988, Changing the data structure from `List` to `Set` as suggested by @quasarchimaere ultimately did the trick.

This PR also contains minor refactoring of the `WonWebsocketHandler`, decoupling websocket/mail/push code from the logic that is responsible for updating the user's atoms.

The bug is hard to reproduce - it seems to depend on the user's data being in a specific configuration; I can reliably get there with one matchat user, but not with any other test user. 

*How to test*:
* check if creating with/without persona works
* go to the detail view of the new posting, check that it is not an external view (eg. the menu would not contain a `Remove` entry)
* Make sure you receive suggestions with that atom.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/java/won/owner/web/websocket/WonWebSocketHandler.java
+++ b/webofneeds/won-owner-webapp/src/main/java/won/owner/web/websocket/WonWebSocketHandler.java
@@ -293,7 +293,8 @@ public class WonWebSocketHandler extends TextWebSocketHandler
             if (!userOpt.isPresent()) {
                 userOpt = Optional.ofNullable(getUserForWonMessage(wonMessage));
             }
-            User user = userOpt.get();
+            User user = userOpt.orElse(null); // it's quite possible that we don't find the user object this way.
+                                              // Methods below can handle that.
             updateUserAtomAssociation(wonMessage, user);
             notifyPerPush(user, atomUri, wonMessage);
             webSocketSessions = findWebSocketSessionsForAtomAndUser(atomUri, user);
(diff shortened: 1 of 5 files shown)


webofneeds/won-owner-webapp/src/main/java/won/owner/web/websocket/WonWebSocketHandler.java
```diff
@@ -369,7 +372,11 @@ private void notifyPerPush(final User user, final URI atomUri, final WonMessage
                     ObjectNode rootNode = mapper.createObjectNode();
                     rootNode.put("type", "HINT");
                     rootNode.put("atomUri", userAtom.getUri().toString());
-                    rootNode.put("connectionUri", wonMessage.getRecipientURI().toString());
+                    if (wonMessage.getRecipientURI() != null) {
+                        rootNode.put("connectionUri", wonMessage.getRecipientURI().toString());
+                    } else {
+                        logger.info("received SocketHint for atom {} without recipientURI", userAtom.getUri());
```
quasarchimaere: please on debug level, unless this is a uncommon logline that indicates a non-exception causing misbehaviour -> then please use the warn level

fkleedorfer: ok: warn.

PR #3014, opened 2019-08-07
Feat  atom card react

quasarchimaere:
Refactors some components to use React instead of angular:
- atom-card
- other-card
- persona-card
- skeleton-card
- atom-card-grid (new component, iterates over atom-cards)
- atom-content-buddies
- atom-content-holds
- atom-content-participants
- atom-content-suggestions
- atom-header
- atom-icon (used to be square-image)
- atom-map (leaflet map -> Caveat: centers to current location if present, if not centers to the first location)
- atom-suggestions-indicator
- labelled-hr
- suggest-atom-picker (addition, react pendant to (still existing) suggestpost-picker -> Caveat: update when entering an uri to fetch does not work that great yet)

This Pull should increase the performance since we are not iterating over many atoms from the state any longer.

Also the leaflet maps (especially in "whats New" and "whatsAround") will not be instantiated if they are not inView (including an offset to render the ones that might be in the view soon if the user scrolls)


fkleedorfer:
Empty inventory of a new user does not look good:

grafik

PR #3059, opened 2019-09-27
DebugMode now also in redux-state. Footer label now updates correctly.

pheara:


quasarchimaere:
i'd prefer if it was in the local component state but now that this is also in the redux state we might be able to do more useful thing with the debugmode present, anyway i think we can change the debugmode check in the ui to be retrieved from the state in general now (just look for the occurences and replace them with the state retrieval) but be aware that we still need the window.debugmode as it is used deep down in the atom creation process

webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
```diff
@@ -573,3 +576,16 @@ export function startTicking() {
...
+function toggleDebugMode() {
+  return dispatch => {
+    won.debugmode = !won.debugmode;
+    dispatch({
+      type: actionTypes.view.toggleDebugMode,
+      payload: won.debugmode,
+    });
+    const text = won.debugmode ? "Debugmode On" : "Debugmode Off";
+    // this.props.toastPush(Immutable.fromJS({ text }));
+    dispatch(actionCreators.toasts__push(Immutable.fromJS({ text })));
```
quasarchimaere: imstead of pushing a toast with a specific message we can use the actionType from the line above to display a toast according to the state within the toast-reducer itself, that way we only have a single dispatch instead of two

m0ru:
TODO only one dispatch

webofneeds/won-owner-webapp/src/main/webapp/app/actions/actions.js
m0ru: done :)

m0ru:
should be mergeable now.

PR #3070, opened 2019-10-08
Some fixes for AtomModelWrapper and Socket

fkleedorfer:
In an attempt to make the randomsimulatorbot usable, found and fixed some bugs

### DIFF
--- a/webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/wonmessage/PrintWonMessageAction.java
+++ b/webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/wonmessage/PrintWonMessageAction.java
@@ -1,0 +1,135 @@
+package won.bot.framework.eventbot.action.impl.wonmessage;
+
+import java.lang.invoke.MethodHandles;
+import java.net.URI;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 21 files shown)


quasarchimaere:
are we sure that the change to a non nullable column as a hibernate annotation is still seen as valid if our flywaydb generated schema is not altered?

fkleedorfer:
Good point

Am 8. Oktober 2019 20:04:20 MESZ schrieb quasarchimaere <notifications@github.com>:

webofneeds/won-bot/src/main/java/won/bot/framework/eventbot/action/impl/wonmessage/PrintWonMessageAction.java
```diff
@@ -0,0 +1,135 @@
...
+        output.append("+-<<< Received " + msg.getMessageType() + " <<<---\n");
+        output.append("| message uri: " + msg.getMessageURI() + "\n");
+        output.append("| sender: " + getSender(msg) + "\n");
+        Optional<String> textMessage = getTextMessage(msg);
+        if (textMessage.isPresent()) {
+            output.append("| message : \n");
+            output.append(format(textMessage.get()));
+        }
+        output.append("+-<<< End of Message <<<---");
+        System.out.println(output.toString());
```
quasarchimaere: can we use the logger instead of System.out.println pls?

PR #3077, opened 2019-10-15
Feat  update 0.6 addtextmessagebehaviour

quasarchimaere:
Creates two new botExtensions:

-ServiceAtomExtension
-TextMessageCommandExtension

These can be seen in action in this (debug-botbranch)[https://github.com/researchstudio-sat/won-debugbot/tree/feat_update_0.6_add_textmsgbehaviour]


sigpie:
Looks good, although Eliza seemed slow to react (small delay before each message). Is this intentional or reproducible?

PR #3081, opened 2019-10-17
Feat  message debugging

fkleedorfer:
AFAICT this fixes the problems we had with the DebugBot. 
Check this PR with debugbot and randomsimulatorbot.


webofneeds/won-bot/src/main/java/won/bot/framework/bot/base/EventBot.java
```diff
@@ -96,8 +106,15 @@ public final void act() throws Exception {
         }
     }
 
+    private void logMessage(WonMessage wonMessage) {
```
quasarchimaere: can be static

PR #3084, opened 2019-10-19
Refactoring WoN messages

fkleedorfer:
Trying to change the approach to messages in WoN.

We used to adhere to the linked data principle for everything, including messages: each message has a `http://` message URI that is dereferencable. For messages, this seems to be a bad idea. The consequence is that the same message, exchanged between two WoN nodes, must have different URIs, or it can only be located on one server. The latter wasn't an option because of other considerations, the former leads into a hell of cross-referencing messages that leads to a very complex system.

In this PR, the attempt is made to change the WoN messages. 
The goals are: 

* [x] a message has exactly one URI (ideally content-addressed), independent of the server they are located on
* [x] strictly immutable messages
* [x] reduce message complexity: only one envelope, no mechanism that a message can contain another. Instead, include message references.
* [x] reduce complexity of message references, which comes for free if messages are content-addressed
* [x] clean up and simplify the message processing on the WoN node: first respond, then forward, then react; 

# Step 1: change message URIs
We introduce `wm` as a new URI scheme for WoN Messages. The format is
`'wm' ':' '/' [message-id] ('/' [path] )* '#' [fragment-id]'`, 
for example, `wm:/1234#content` may identify a content graph in message 1234.
For dereferencing, one must know which WoN node the message is located on, and then replace the prefix `wm:/` with the event URI prefix of the node, for example, `https://node.matchat.org/won/resource/event/1234#content`.

Seems done with [be74c](https://github.com/researchstudio-sat/webofneeds/commit/be74cfc6b7ff8b4ed0969d8b7e740892af102cd2). 

# Step 2: Same message in both containers

The goal here is to send a message on to the recipient node without adding an envelope or making a 'remote message'. The SuccessResponses are supposed to work the same way. Small extension here: the sender node's SuccessResponse should be sent along with the original message to the recipient node so that both message containers have exactly the same messages.
Todo:

* [x] set up a unit test fixture for refactoring
* [x] change DB to have an M:N relationship between message and container
* [x] check that everything still works
* [x] send same message as outgoing in a test
   Now, the consequences start showing:
       + we cannot wrap an envelope around a message to add data. That data has to be added to a response
           - connect: the connection URI is generated on the node and added to the outer envelope, so the receiving node can get the connection URI from the message. That won't work now. We have to wait for the successResponse of the sender node to get the connectionURI (which is the sender).
           - looking at options, realized how confusing the OPEN message is - got rid of that 63e4c0d

* [x] adapt checks for incoming message
* [x] adapt message handling code to work without 'fromExternal' and 'CorrespondingRemoteMessage'
* [x] rework node message handling to new message structure
* [x] adapt WonMessage so it can hold one or more messages (message + response1 + response2 + forwarded1 + ... + forwardedN)
* [x] fix WonMessageBuilder so it works like a normal builder
* [x] new connection URIs as sub-paths of atom URIs
* [x] change message referencing code to reference only finished delivery chains
* [ ] change node logic to delete failed delivery chains
# Webapp TODOs
Much of the information below has been written up in the `messages.md` and `delivery-chains.md` files in this folder in PR #3120: https://github.com/researchstudio-sat/webofneeds/tree/message-refactoring-docs/documentation/protocol
* [ ] remove all traces of OPEN messages. Use CONNECT instead. 
* [ ] replace all usages of `msg:receivedTimestamp` and `msg:sentTimestamp` with `msg:timestamp`
* [ ] eventually we'll have less info in a message, we'll need more logic to determine senderAtom/recipientAtom/recipient connection etc. Have a look at `won.protocol.service.impl.MessageRoutingInfoServiceImpl` for inspiration
* [ ] both responses (from the sender node and the recipient node) refer  to the original message via `msg:isResponseTo` - `msg:isRemoteResponseTo` is dropped. It consititutes redundant information and thus raises the chance for programming errors.
* [ ] `msg:correspondingRemoteMessage` is dropped. The concept does not exist any more
* [ ] `msg:isResponseTo` is renamed to `msg:respondingTo` 
* [ ] `msg:isResponseToMessageType` is renamed to `msg:respondingToMessageType`
* [ ] atom content can no longer contain uris that are sub-paths of the atom URI, i.e. `[atomUri]/location/geo` etc. are no longer allowed
* [ ] a failure response is no longer stored by the node that generates it. A local failure response is delivered to all owner apps that are registered for the atom. The original message is not stored. A failure response from a remote node will be delivered to all owner apps as well, and then, the complete delivery chain is deleted - including the message and the local response.
* [ ] An **echo** can no longer be identified by its outer envelope being FROM_OWNER as opposed to FROM_EXTERNAL for an incoming, non-echo message  - it's the same message. Rather, when delivering an echo, the RDF dataset contains the response message generated by the sender's node in addition to the original (=echoed) message.
* [ ] An **incoming message** is now accompanied by the two responses (one of the sender's node and one of the recipient's node) in the same RDF dataset.
* [ ] The **connection URI** is now `[atomUri] + '/c/' + [connectionId]` - because we want to reduce round trips and it's an unnecessary round trip to find out which atom a connection belongs to if we can encode it in the URI.
* [ ] The **connection container URI** is now `[atomUri] + '/c' `, which makes more sense in connection with the previous change than the `[atomUri] + '/connections'`.
* [ ] message containers have been renamed from `[...]/events` to `[...]/msg`
* [ ] A hint does not have a `won:senderNode` property any more
* [ ] A hintFeedbackMessage addresses the connection using `won:connection`, (`won:sender` and `won:recipient` are dropped)
* [ ] each message now has one envelope graph and one signature graph. The names of these graphs in the mesage dataset are URIs of the form `[msgUri]#envelope` and `[msgUri]#signature`. The signature graph contains exactly one signature, which refers to the graphs it signs via `msg:signedGraph`. These graphs must be found in the messsage. Most message types have just this one signature, signing all graphs. The exception are CREATE_ATOM and REPLACE messages. There, each content graph has a separate signature in the envelope graph. Thus, the content can be verified against its signature separately from the message.
* [ ] `resumeAfter` and `resumeBefore` used in our Linked Data Paging containers now work on complete URIs, not just connection/atom/message Ids.
* [ ] When composing a message, the message URI **must** be `wm:/SELF`. It will be signed with this URI, then `SELF`it is replaced by the hash of the message.
* [ ] The `sig:` prefix and corresponding namespace is no longer to be used
* [ ] The envelopeGraph is no longer linked to the message URI via `rdfg:subGraphOf`. Now it's [messageURI] msg:envelope [envelopeGraphUri], same approach for the signature graph using `msg:signature`.


webofneeds/won-core/src/main/java/won/protocol/util/linkeddata/CachingLinkedDataSource.java
```diff
@@ -242,9 +255,11 @@ private DatasetResponseWithStatusCodeAndHeaders fetchOnlyOnce(final URI resource
         CountDownLatch preExistingLatch = countDownLatchMap.putIfAbsent(cacheKey, latch);
         try {
             if (preExistingLatch != null) {
-                logger.debug("resource " + cacheKey
-                                + " is being fetched in another thread, we wait for its result and use it "
-                                + "if it turns out to be cacheable");
+                if (logger.isDebugEnabled()) {
+                    logger.debug("resource " + cacheKey
```
quasarchimaere: not to nitpick, but if you use a parameterized query ( .. {}... with the cacheKey as a parameter) you could avoid the isDebugEnabled clause, because the logger framework would only concat/init the log-string if the debug mode is enabled anyway

webofneeds/won-core/src/main/java/won/protocol/util/linkeddata/CachingLinkedDataSource.java
```diff
@@ -292,6 +311,9 @@ private DatasetResponseWithStatusCodeAndHeaders fetchAndCacheIfAppropriate(final
             if (expires != null && expires.getTime() == 0) {
                 // the expires header was invalid (e.g. '0'), which means: already expired.
                 // Don't cache.
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Fetched {}. Will not be cached due to invalid 'Expires' header sent by server");
```
quasarchimaere: missing a parameter (and if clause is not necessary as long as the param wouldnt be calculated within the log statement itself)

webofneeds/won-core/src/main/java/won/protocol/util/linkeddata/CachingLinkedDataSource.java
fkleedorfer: fixed, thanks

webofneeds/won-core/src/main/java/won/protocol/util/linkeddata/CachingLinkedDataSource.java
fkleedorfer: fixed, thanks

webofneeds/won-owner-webapp/src/main/java/won/owner/pojo/RestStatusResponse.java
```diff
@@ -38,7 +38,10 @@
     EXPORT_NOT_VERIFIED(7403, "You need a verified account to export your data", HttpStatus.BAD_REQUEST),
     RECOVERY_KEYGEN_USER_NOT_FOUND(8100, "Could not generate recovery key: user not found", HttpStatus.BAD_REQUEST),
     RECOVERY_KEYGEN_WRONG_PASSWORD(8101, "Could not generate recovery key: wrong password", HttpStatus.BAD_REQUEST),
-    SUBSCRIBE_SUCCESS(8200, "Successfully subscribed to push notifications", HttpStatus.OK);
+    SUBSCRIBE_SUCCESS(8200, "Successfully subscribed to push notifications", HttpStatus.OK),
+    ERROR_SENDING_MESSAGE_TO_NODE(9000, "Error sending WoN message to node", HttpStatus.INTERNAL_SERVER_ERROR),
+    ERROR_PREPARING_MESSAGE(9000, "Error signing and sealing WoN message", HttpStatus.INTERNAL_SERVER_ERROR),
```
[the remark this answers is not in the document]
fkleedorfer: Fixed

PR #3087, opened 2019-10-21
Update elm to 0.19.1

### LINKED ISSUE #2901: PublishButton "crashes" when being unloaded
When the PublishButton elm component is unloaded, the `getElement` call fails which, when running in optimized mode, causes [a known bug in Elm](https://github.com/elm/compiler/issues/1836) to occur.

This shows up in our application as something like
> TypeError: QA is not a function

where `QA` could change to something else as our code changes.

A fix for this is on the way and will land in Elm 0.19.1 "soon"

The current impact is just a stray error message, since our application doesn't actually care about the call succeeding.

### LINKED ISSUE #3066: Switch npm install to npm ci
`npm install` has a habit of changing small details in the `package-lock.json` on different machines if run, which makes `package-lock.json` unstable across environments.
npm introduced the `ci` subcommand to fix this, which does not change the `package-lock.json` under any circumstances, and throws an error if there is an incompatibility with `package.json`.

Currently blocked by https://github.com/eirslett/frontend-maven-plugin/issues/855 in the maven plugin we use to interface with npm

### LINKED ISSUE #3072: Change owner-webapp build back to webpack
**Is your feature request related to a problem? Please describe.**
parcel bundler is unable to bundle our webapp on windows environments, since we our webapp is developed within unix and windows environments it is essential that it also bundles in both environments

**Describe the solution you'd like**
Change the bundle from parcel back to webpack

**Describe alternatives you've considered**
lets not waste more time on alternatives

**Additional context**
none

peacememories:
This should hopefully fix the problems with concurrent builds on Windows.
It should also theoretically result in Elm debuggers flying around everywhere, but this doesn't seem to be happening. Please test whether that is the case for you.

## Testing
It's very important to fully clear `node_modules` before testing this, because we previously applied patches to a package, which should now be reverted (or not applied at all, ideally)

Closes #3066 
Closes #3072 if everything works now
Fixes #2901

Please do not merge without talking to me first to make sure everything works as expected.


maxstolze:
- [ ] Tested on Windows with Eclipse, woks

quasarchimaere:
build also runs at work, i think we can merge this but i will wait for @peacememories approval/inspection

PR #3130, opened 2019-12-05
Fix  ui proposals

quasarchimaere:
Fixes displaying sent/received proposals or anything that causes messageEffects to work without a reload


fkleedorfer:
Kind-of works but more complex workflows like propose-accept-proposetocancel-accept do not lead to the expected outcome

quasarchimaere:
still probably better than a necessary reload when a proposal comes in

PR #3245, opened 2020-04-15
Added handler for clicking on map

### LINKED ISSUE #3091: Location can't be selected by clicking on the map
**Describe the bug**
This is a regression, as it was already possibly to select a location with a single click on the map before. Expected behaviour here is up for discussion as it relates to #2366 

If this was intentionally removed, this issue and #2366 should be closed.

peacememories:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Build was run locally and `mvn install` succeeds

## Pull request type

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

- Fixes #3091 

## What is the new behavior?

- When clicking on the map in the location picker, a marker is placed at the selected position and the location bar is updated with the corrdinates of the marker.
## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Having the coordinates in the location bar is not very elegant, especially since right now they can't be entered that way. An alternative would be to use nominatim to reverse-search an address, but trying that I had very inconsistent results.

A third option would be to use the original lat and lng, but use the location name provided by nominatim. I am not sure how this would affect later matching of the created atoms though.

### DIFF
--- a/webofneeds/won-owner-webapp/src/main/webapp/app/components/details/picker/location-picker.jsx
+++ b/webofneeds/won-owner-webapp/src/main/webapp/app/components/details/picker/location-picker.jsx
@@ -123,7 +123,11 @@ export default class WonLocationPicker extends React.Component {
 
     if (selectedLocationCoordinates) {
       selectedLocationMarker = (
-        <Marker position={selectedLocationCoordinates} icon={locationIcon} />
+        <Marker
+          position={selectedLocationCoordinates}
+          icon={locationIcon}
+          onClick={() => false}
+        />
       );
     }
 
@@ -184,6 +188,7 @@ export default class WonLocationPicker extends React.Component {
         <Map
           className="lp__mapmount"
           zoom={13}
+          onClick={e => this.mapClicked(e)}
           center={
             selectedLocationCoordinates ||
             currentLocationCoordinates || [48.210033, 16.363449]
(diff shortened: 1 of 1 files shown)


fkleedorfer:
That works. However, we cannot post the resulting atom. The console says:
```
jsonld-utils.js:120 Uncaught (in promise) TypeError: Cannot read property 'lat' of undefined
    at Object.s [as parseSPlace] (jsonld-utils.js:120)
    at Object.parseFromRDF (location.js:35)
    at b (parse-atom.js:296)
    at m (parse-atom.js:53)
    at s (reduce-atoms.js:9)
    at y (atom-reducer-main.js:158)
    at n (reducers.js:251)
    at reducers.js:253
    at immutable.js:1377
    at Yt.iterate (immutable.js:1744)
```

quasarchimaere:
the issue stems from the (faulty) structure of the location that is created by the location-picker when clicking, `nwCorner` and `seCorner` aka the locations boundingBox is still mandatory for any location, this is not an issue when selecting a location via the textbox, because the helper function `nominatim2draftLocation` seen in the file `nominatim-api.js` creates a valid location... long story short to fix this problem please inject the correct structure for the location json object

`
lat: val,
lng: val,
name: val,
nwCorner: {
      lat: val,
      lng: val,
    },
    seCorner: {
      lat: val,
      lng: Number.parseFloat(b[3]),
    },`

PR #3246, opened 2020-04-15
Update agreement tests and refactor AgreementProtocolState

maxstolze:
1. Update agreement tests for the refactored message handling
2. Adding "agreedMessageUri" to AgreementProtocolState to handle in webapp
3. Refactor CancelationPendingProposals without using the DefaultModel of pendingProposals
4. Recalculate effects directly from explizit MessageEffects for claims, proposals, proposeToCancel, agreedMessages


fkleedorfer:
Trying this PR via the webapp:
1. create atom with user A
2. create atom with user B
3. connect from A, accept from B
4. send a message with B and propose it
5. click on "show agreement data" shows the proposal for A as well as B
6. accept the proposal with A

Now *do not reload either window*:
* in A's conversation view, the proposal is changed into an agreement but not in B's
* in A's conversation view, the agreement is greyed out, but the 'propose to cancel button still works'
   - why is it greyed out?
   - the fact that it is greyed out suggests it is somehow deactivated, which clashes with the working button
* in A's 'show agreement data', we see the agreement
* in B's 'show agreement data', we it's still only a proposal
* from B's view, we can still retract the proposal, and when we do, the 'show agreement data' view is empty
* when reloading after this, both sides show the agreement in 'show agreement data', but with strikethrough and greyed out

fkleedorfer:
... Instead, if you reload after step 6:
* we have the agreement in both windows
7. let B propose to cancel
8. let A accept that
* now, even after a reload, B can still retract their cancellation proposal
* in B's 'show agreement data' view, the cancellation proposal is still unaccepted

webofneeds/won-utils/won-utils-conversation/src/main/java/won/protocol/agreement/AgreementProtocolState.java
```diff
@@ -103,18 +98,14 @@ public AgreementProtocolUris getAgreementProtocolUris() {
                 ProposalUris proposal = new ProposalUris(m.getMessageURI(), m.getSenderAtomURI());
                 proposal.addProposes(proposed);
                 proposal.addProposesToCancel(cancelled);
+                // TODO
```
fkleedorfer, line 101: TODO?

webofneeds/won-utils/won-utils-conversation/src/main/java/won/protocol/agreement/AgreementProtocolState.java
```diff
@@ -83,18 +76,20 @@ public AgreementProtocolUris getAgreementProtocolUris() {
...
             boolean isProposal = false;
             if (!cancelled.isEmpty()) {
                 // remember this is a pending proposal that would cancel stuff
-                uris.addPendingCancellationProposalUri(m.getMessageURI());
+                // uris.addPendingCancellationProposalUri(m.getMessageURI());
                 isProposal = true;
             }
             if (!proposed.isEmpty()) {
                 // remember this is a pending proposal that proposes stuff
+                // TODO
```
fkleedorfer, line 92: TODO?

webofneeds/won-utils/won-utils-conversation/src/main/java/won/protocol/agreement/AgreementProtocolState.java
```diff
@@ -83,18 +76,20 @@ public AgreementProtocolUris getAgreementProtocolUris() {
...
+            // TODO
             uris.addCancellationPendingAgreementUris(cancelled);
             // determine what it proposes
             Set<URI> proposed = m.getEffects().stream().filter(e -> e.isProposes()).map(e -> e.asProposes())
                             .flatMap(e -> e.getProposes().stream()).collect(Collectors.toSet());
             boolean isProposal = false;
             if (!cancelled.isEmpty()) {
                 // remember this is a pending proposal that would cancel stuff
-                uris.addPendingCancellationProposalUri(m.getMessageURI());
+                // uris.addPendingCancellationProposalUri(m.getMessageURI());
```
fkleedorfer, line 87: Dead code

webofneeds/won-utils/won-utils-conversation/src/main/java/won/protocol/agreement/AgreementProtocolState.java
```diff
@@ -83,18 +76,20 @@ public AgreementProtocolUris getAgreementProtocolUris() {
             Set<URI> cancelled = m.getEffects().stream().filter(e -> e.isProposes()).map(e -> e.asProposes())
                             .flatMap(e -> e.getProposesToCancel().stream()).filter(this::isAgreement)
                             .collect(Collectors.toSet());
+            // TODO
```
fkleedorfer, line 79: TODO?

webofneeds/won-utils/won-utils-conversation/src/main/java/won/protocol/agreement/AgreementProtocolState.java
```diff
@@ -50,9 +43,8 @@
     private final Dataset cancelledAgreements = DatasetFactory.createGeneral();
     private final Dataset rejected = DatasetFactory.createGeneral();
     private Dataset conversation = null;
-    private final Set<URI> retractedUris = new HashSet<URI>();
-    private final Set<URI> claimedUris = new HashSet<URI>();
-    private final Set<URI> acceptedCancellationProposalUris = new HashSet<URI>();
+    private final Set<URI> retractedUris = new HashSet<URI>(); // TODO: Check!
```
fkleedorfer, line 46: What are these TODOs?

